PR #10, opened 2015-04-29
Update from template

gvwilson:


r-gaia-cs:
@gvwilson this pull request brings all the history of template. I'm not sure if this is what we want.

gvwilson:
Hm - OK, what should we do instead?

On 2015-04-30 7:59 AM, Raniere Silva wrote:
>
> @gvwilson <https://github.com/gvwilson> this pull request brings all 
> the history of template. I'm not sure if this is what we want.
>
> —
> Reply to this email directly or view it on GitHub 
> <https://github.com/swcarpentry/instructor-training/pull/10#issuecomment-97748590>.
>

-- 
Dr. Greg Wilson    | gvwilson@software-carpentry.org
Software Carpentry | http://software-carpentry.org

r-gaia-cs:
> what should we do instead?

I will check tonight.

andreabedini:
I think the right way is to rebase this PR squashing all commits with `git rebase -i`. Not entirely sure thought.

wking:
On Thu, Apr 30, 2015 at 04:59:22AM -0700, Raniere Silva wrote:
> @gvwilson this pull request brings all the history of template. I'm
> not sure if this is what we want.

I think this *is* what you want, since having that history:

a. makes it easier for the local maintainer to understand why the
   template has changed, and
b. makes it easier for Git to merge subsequent template changes,
   because Git will be able to find common ancestor commits.

r-gaia-cs:
@gvwilson The long list of commits is because is the first time that https://github.com/swcarpentry/styles/ is included here. +1 to merge.

PR #34, opened 2015-12-03
Mentioning the J word

### LINKED ISSUE #33: Don't say "just"
In the "Things You Shouldn't Do in a Workshop" section, I think it would be good to add that instructors shouldn't tell participants to "just" do something (e.g., "just go to the command line and type..."). This was brought up in the instructor training I participated in, and I thought it was a really important point. If someone says "just do X" and you can't do it, that's pretty demoralizing.

gvwilson:
1.  Adding forward mention from 03-maps about the J word.
2.  Adding note to "Things you Shouldn't Do" about the J word with backward ref.

This closes #33.


embray:
:+1:

PR #42, opened 2015-12-07
fixing transposition of link in markdown

jiwalker-usgs:
I'm taking the instructor training and noticed this broken link


gvwilson:
Thanks for the catch.

Issue #30, opened 2015-11-30
Extract text of exercises from Piazza


gvwilson:
Get both exercise descriptions and typical student responses from Piazza and fold into material (the latter will be good examples to show learners).  This replaces point 2 of #19.

Issue #29, opened 2015-11-30
Flag challenges as in-person or online


gvwilson:
Replaces item 1 of #19.

PR #43, opened 2015-12-08
presenting one way of making these materials more out-of-the-box.

ChristinaLK:
Hi everyone - I wanted to capture some of this material while the ideas are fresh!  Some things that are missing for me when working from these materials are: 

* discussion questions, like the questions Greg threw into the chat window during online training, and otherwise can be used either to check-in (formative assessment) or to spur discussion/prime for an idea
* pre-made notes and/or assignments that I could just cut and paste into the etherpad.  The assignments especially can't just be cut and pasted from the current website as is.  

So I've added some material to the repo (for just one topic!) as a place to start discussing (this is not a PR to be merged, necessarily!): do we include this kind of material?  If so, how?  Etc.  All for your consideration.   :)  

The major downside I see of this model is maintaining multiple copies of things - every time you edit an exercise, you have to change the plain text version too.  But someone is going to have to do that work anyway when the exercises are used, so...

### DIFF
--- a/02-models.md
+++ b/02-models.md
@@ -15,6 +15,9 @@ subtitle: Mental Models
 > It's what you know for sure that just ain't so.
 > - Mark Twain
 
+> Discussion question
+> What do we mean when we say that someone is a novice? What are the symptons of being a novice, and what are the cognitive causes?
+
 Our goal is not to teach the syntax of a particular programming language,
 but to teach people how to think about programming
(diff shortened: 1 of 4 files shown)


ChristinaLK:
Oops - didn't realize I was addressing own issues I'd raised earlier #29 #30 .  Consistency!

ChristinaLK:
not necessary, due to changes in #67

PR #45, opened 2015-12-11
Notes for instructor's guide

gvwilson:
For review and additions.


ChristinaLK:
This looks great - I'm not thinking about this in detail right now, but I will probably revisit next week or the week after.

PR #50, opened 2016-01-07
Adding detailed trainers guide for 2day f2f workshop

apawlik:
This schedule looks very detailed and rigid but should help new trainers work out how to plan the 2day f2f workshop and split the material.

This is only an initial stub. I'd suggest to split trainer's guide as it is now (http://swcarpentry.github.io/instructor-training/instructors.html) into separate guides for different forms of deliver (multi-week online; 2day online and f2f online).

### DIFF
--- a/trainer-guide-2-day.md
+++ b/trainer-guide-2-day.md
@@ -1,6 +1,6 @@
 ---
 layout: page
-subtitle: Sample workshop guide for trainers
+subtitle: 2-day face-to-face Instructor Training guide for Trainers
 ---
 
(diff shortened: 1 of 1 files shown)


gvwilson:
Thanks for the schedule - can you please amend after Lausanne and 
Australia, and then we'll merge?

apawlik:
Yup. Sure.

ChristinaLK:
I love this!  Thanks for putting it together @apawlik.  :)

I can contribute more detailed thoughts (I frame the first / second day slightly differently), but will wait per Greg's comment above.

gvwilson:
I'd prefer to keep all this material in instructors.md, and to remove the bold face, but the content is +1 to merge - thanks!

PR #64, opened 2016-01-11
Added ref to imposter study among highly achieving women, 1978

apawlik:
Apparently this is one of the first studies on imposter syndrome - also among highly achieving women.


gvwilson:
Added manually in 2d067be05507 because of filename changes.

PR #66, opened 2016-01-13
added before-the-workshop email from Greg

tracykteal:
Added the 'before the workshop' email, that instructors can modify for their workshop and send out, that details what people need to do before they arrive.


tracykteal:
This PR included more than it should have. Closing and resubmitting correct one.

PR #87, opened 2016-02-03
The language in this section implied that readings are required prepa…

arokem:
…ration.

Accordingly: moved them to the front page into a "preparation" box.

Also: since they are three, the word "both" was confusing.


02-terms.md
```diff
@@ -188,7 +188,7 @@ Shorter pieces include
 Guzdial's "[Why Programming is Hard to Teach](papers/guzdial-why-hard-to-teach-2011.pdf)"
 and "[Top 10 Myths About Teaching Computer Science](papers/guzdial-10-myths-2015.pdf)",
 and Porter et al's "[Success in Introductory Programming: What Works?](papers/porter-what-works-2013.pdf)",
-both of which should be read before starting this class.
+all of which you should have read before starting this class.
```
gvwilson: all of which you should read before starting this class. (?)

arokem:
Previous comment addressed + rebased.

PR #90, opened 2016-02-15
Memory

arokem:
Suggestion for an online exercise/demonstration of memory serial position effects, which demonstrate the division of memory into long-term and short-term effects. 

The serial position effect itself is also important when considering lesson design, because (depending on many other factors) participants are prone to remember facts presented in the beginning and end of a lesson more clearly than the middle.

### DIFF
--- a/06-memory.md
+++ b/06-memory.md
@@ -24,11 +24,11 @@ limits of human mind and memory, and how that may impact learning.
 
 ## Plus or Minus Two
 
-> ## Faces and Names
+> ## The serial position effect
 >
-> Add an exercise here where people are shown a bunch of faces/names
-> (or similar list of random facts) and
-> then have to name as many as they can remember?  
+> Go to [this website](http://cat.xula.edu/thinker/memory/working/serial) and
+> follow the instructions therein.
+>
 {: .challenge}
 
 While the graph model of knowledge is inaccurate but useful,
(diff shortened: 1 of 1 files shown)


wking:
On Mon, Feb 15, 2016 at 01:40:33PM -0800, Ariel Rokem wrote:
> Suggestion for an online exercise/demonstration of memory serial
> position effects…

I'd recommend linking [1] as well, since [2] shows me “A plugin is
needed to display this content”.

“Go here and follow instructions to learn something that I don't
summarize here” is overly mysterious.  I'd recommend some version of
Wikipedia's opener as a summary, so folks who don't question that
finding can skip the Xavier applet and Wikipedia article.

[1]: https://en.wikipedia.org/wiki/Serial_position_effect
[2]: http://cat.xula.edu/thinker/memory/working/serial

arokem:
Thanks for the comment. I have refactored the challenge to an "analog" version (the list is there for people to read). I have also added the link to the Wikipedia article in a paragraph that explains this effect.

wking:
On Mon, Feb 15, 2016 at 03:14:27PM -0800, Ariel Rokem wrote:
> I have refactored the challenge to an "analog" version (the list is
> there for people to read). I have also added the link to the
> Wikipedia article in a paragraph that explains this effect.

Looks good to me, thanks :).

ChristinaLK:
this is a great addition - thanks for turning my rather vague exercise suggestion into something real and science-y.  :)

gvwilson:
We're all about the science-y.

PR #117, opened 2016-03-04
Link to the PDF of the Parnas/Clements article, instead of the ACM page.

arokem:
The ACM page doesn't have a PDF, and someone might (erroneously) think that no 
PDF is available.

### DIFF
--- a/07-design.md
+++ b/07-design.md
@@ -106,7 +106,10 @@ Wiggins and McTighe use the acronym WHERETO:
 > ## How and Why to Fake It
 >
 > One of the most influential papers in the history of software engineering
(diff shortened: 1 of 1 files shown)


wking:
On Fri, Mar 04, 2016 at 05:42:03AM -0800, Ariel Rokem wrote:
> The ACM page doesn't have a PDF, and someone might (erroneously)
> think that no PDF is available.

I think it's probably best to link the DOI [1], and then (if you want)
list a PDF parenthetically.  For example:

  [A Rational Design Process: How and Why to Fake
  It](http://dx.doi.org/10.1109/TSE.1986.6312940)
  ([PDF](http://www.ics.uci.edu/~taylor/classes/121/IEEE86_Parnas_Clement.pdf))

That way something survives Prof. Taylor's site going down.

[1]: http://dx.doi.org/10.1109/TSE.1986.6312940

arokem:
Thanks for the suggestion. Changed accordingly.

Issue #83, opened 2016-01-29
Feedback from Lausanne


gvwilson:
1. More time for video exercises. Also provide them guidance about how much time they should spend on each feedback. So for example: each video 2 minutes and each feedback about 3 min. I think they were also running out of time because we didn't tell them how long to give feedback on one video. Giving them just the grid for feedback is not enough to structure it.

2. Theming the exercises on Day 1 (so build MCQ, concept map and then Learning Objectives) was something that people really liked.

3. More time for the video on Day 2 when they record themselves live coding. Also give them better guidance for preparations that we want them to pick something from the middle of the materials so that they actually spend most time doing live coding than introducing the topic. This should actually be really well demonstrated during the demo (i.e. how to teach stuff taken out of context and make assumptions that for example people already opened their terminals).

4. During the live coding demo, Tracy purposely made some common mistakes - this worked well when collecting feedback (we did one up one down).

5. A very high level comment: when talking about the concepts/theory the trainer needs to make very clear and probably a good few references back to how it ties in with SWC/DC. Sure, yeah, this is supposed to be generic part of the training. However, people come in for SWC/DC and they want to know "Why are you teaching us about cognitive overload? Is this used in SWC/DC". This basically boils down to the trainer being very well read in the pedagogical topics we cover and also be an experiences instructor.

6. People would like to get some kind of a list of the main concepts. Personally I (Aleksandra) think that we could have in the repo a simple print out sheet which only has a list of these key concepts ("Summative assessment. Formative assessment. Mental model. Concept map. COngnitive overload. …….) with spaces left between them and print them and hand them out. People can take notes on them. Taking notes in Etherpad is not great - it becomes messy. People are taking notes in their notepads anyway but they need a highlight/signpost - we say many words during the training but they sometimes are not sure which were the key ones.

7. The part on "Materials and how to contribute" needs better structure. Open discussion is rubbish. EIther we decide we focus on the "meta" level - talk about what we need in terms of materials, the flow in the community (maintainers, contributors, small changes vs large overhauls etc). or we discuss the mechanics (PRs). Personally I think the latter, the risk is that people in the IT who know Git well get bored. We should rather think about having a support group for people who want to learn Git too be able to do PRs to materials and then explain trainees how they can get into contributing to materials.
I would also cover more about how friendly the community is - showing examples of good discussions under PRs (possibly controversial PRs). I also showed how the thread on Discuss list about "Leaving novices behind" turned into blog post.

[...]

PR #145, opened 2016-03-22
adding beginning of outline for trainees to use during sessions

ChristinaLK:
Still needs revision - comments welcome!

Addressing point 6 in issue #83

### DIFF
--- a/messages/outline.md
+++ b/messages/outline.md
@@ -1,5 +1,5 @@
-Terms
-=====
+General Terms
+=============
 
 * Educational psychology
 	* Cognitivist
@@ -7,7 +7,7 @@ Terms
 * Instructional Design
 * Pedagogical Content Knowledge
 
-Theory + Tools
+Theory and Tools
 ==============
 
 Cognitive Development + Mental Models
@@ -68,14 +68,11 @@ Performance and Practice
 
 * What are advantages / disadvantages to live coding? 
 
-* What other practices are used in SWC/DC workshops?  
+* What other practices are used in Software / Data Carpentry workshops?  
 
 * What factors will: 
 	* motivate learners?  
 	* demotivate learners?
-		* stereotype threat
-		* imposter syndrome
-		* mindset
 
 Community of Practice
 =====================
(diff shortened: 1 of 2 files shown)


gvwilson:
Is this consistent with the revised schedule?  Do you see this being pasted into the Etherpad at the start of training, or should it be incorporated into messages/welcome.txt?

ChristinaLK:
I organized by topic, not schedule, as I was thinking of this more as `messages/outline.txt` document, than something to be pasted into the etherpad.  One of the points raised by @apawlik was the potential need for a notes platform besides the etherpad.   

It's also a nice "meta" for the idea of handouts in an actual SWC workshop.  :)

ChristinaLK:
Can @apawlik take a look at this?  Was this what you had in mind, in terms of a handout?

apawlik:
Thanks for this. Looks good - we may add a few more things in due course.

Yes, I thought about something like this in terms of a hand-out. Based on the feedback that I get and the observation from lots of workshops, I will strongly advocate for a handout that could be printed out and given to people. **Lots** of people take notes during the IT in their notebooks and often say that having a  handout on which they could take notes would be useful.
**It doesn't matter** that we have materials online. People want a handout. I think the fact that *some* of us (trainers) rely 100% on electronic notes doesn't mean we have to force trainees to do so.
Etherpad is also not great - it's a MESS half-way through the workshop.

ChristinaLK:
@apawlik, I agree that a handout like this (that people can fill in + annotate) is a good idea.  Ideally then, this should be a pdf.  Suggestions for how to make that happen?  @gvwilson?

Also, Aleksandra, any comments on format or content?  I tried to pull major terms from each section, but didn't explicitly list things like "faded examples" or "multiple choice questions" - I thought people could write them in under the "teaching tools" item in each section.  Does that make sense?  Do you think learners will think that way?  Does it matter?

gvwilson:
If y'all can create the content, I can take care of PDF'ing.

ChristinaLK:
Finally picking this back up again - I moved into `messages`.  If it can be pdf'd, that'd be great.  I'll definitely try to use this in June + let you know how it goes.

PR #149, opened 2016-03-22
Expanding note on I/We/You

### LINKED ISSUE #71: I/we/you
In SWC we present materials as I/we/you. I introduce, we do it together, you practice on your own. Add this info to the instructor notes.

Another way of talking about it is Guided Practice and Independent Practice.

gvwilson:
Closes #71

### DIFF
--- a/08-practices.md
+++ b/08-practices.md
@@ -29,39 +29,22 @@ and work through the lesson,
 typing in the code,
 reformatting data,
 and talking as we go.
-This is an example of *I/we/you* teaching:
-the instructor does it,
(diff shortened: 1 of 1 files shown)


ChristinaLK:
This information is found in the section on "lesson study."  So I might recast as: 

"as discussed in (link to lesson study), this is an example of I/We/You teaching, where...[describe briefly].  In this case, we think it's a better approach than you/y'all we, because [explain...]."  

Does that make sense?  I like the highlighting back to previous info, but maybe it's okay to just present again in a different context...

PR #152, opened 2016-03-26
Temporary fix for CSS used on print

rgaiacs:
# Before

software carpentry_1

software carpentry_2

software carpentry_3

# After

training1

training2

# Details

This remove the header and footer on the print.

### DIFF
--- a/_sass/foundation-components/_type.scss
+++ b/_sass/foundation-components/_type.scss
@@ -472,10 +472,15 @@ $align-class-breakpoints:
       .print-only { display: none !important; }
       @media print {
         * {
-          background: transparent !important;
           color: $black !important; /* Black prints faster: h5bp.com/s */
           box-shadow: none !important;
           text-shadow: none !important;
+        }
+
+        #navigation,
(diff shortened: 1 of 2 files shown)


_sass/foundation-components/_type.scss
```diff
@@ -472,12 +472,17 @@ $align-class-breakpoints:
       .print-only { display: none !important; }
       @media print {
         * {
-          background: transparent !important;
```
rgaiacs: @gvwilson This is why the logo doesn't show up. The logo is white and this makes the background also white.

_sass/foundation-components/_type.scss
```diff
@@ -472,12 +472,17 @@ $align-class-breakpoints:
...
           color: $black !important; /* Black prints faster: h5bp.com/s */
           box-shadow: none !important;
           text-shadow: none !important;
         }
 
+        #navigation,
+        #masthead-no-image-header,
```
rgaiacs: @gvwilson The header need some fix to the size. This just disable the header.

PR #189, opened 2016-05-13
Add a missing closing parenthesis in Motivation and Demotivation

cmd-ntrf:
When refering to Margolis and Fisher, the parenthesis was left open, leaving the reader to think that the under-representation of racial group was actually part of the reference.


gvwilson:
Good catch - thanks.

PR #215, opened 2016-06-07
14-carpentries.md: Remove extra '?' in 'questions'

iglpdc:
I think it's a typo, but maybe you're trying to be emphatic.


gvwilson:
Good catch.

PR #216, opened 2016-06-07
Update checkout.md

raynamharris:
Added the new pick one of three options for contributing to our lessons. 

I'm not sure how much of the information about proposing new changes should stay and what should go.  Any suggestions before I start cuttings?

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -11,7 +11,7 @@ After you have finished the two-day training class,
 you must go through these steps in order to complete your training and be
 fully certified as a Data and/or Software Carpentry instructor:
 
-1.  Submit a [small change to one of our lessons](#lesson-change).
+1.  Make a [contribution to a lesson's content, exercises, or instructor's guide](#lesson-change).
 2.  Take part in an online [discussion session](#discussion-session).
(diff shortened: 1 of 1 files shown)


gvwilson:
thanks

ErinBecker:
Thanks Rayna. Just a quick question. Is the following meant to apply in both directions (i.e. can DC instructors qualify by submitting a change to SWC lesson?)?

>Note that you may submit a change to a Data Carpentry lesson when qualifying as a Software Carpentry instructor: the goal is to be sure you are familiar with the process.

I assume the answer is yes. If so, can we add a "and vice versa" just to stave off confusion?

Thanks again!

raynamharris:
@ErinBecker yes, the wording should be vice versa, to the best of my knowledge

ErinBecker:
This contradicts what Greg said in training today, so lets make sure all
are on the same page.

What was said in training was that DC instructors need to contribute to a
DC lesson and do lesson demo on DC materials. Vice versa for swc
instructors and that anyone who wanted to get double certified needs to do
both of those twice.

Would be good to clarify so we are all communicating the same thing.

raynamharris:
Thanks @ErinBecker . See new PR #225

PR #218, opened 2016-06-10
Moved challenges up to earlier in the lesson

raynamharris:


ChristinaLK:
+1 to merge - I'll let someone else take a look before I do though.  :)

lexnederbragt:
A quick look indicates that perhaps the callout indicator(s) didn't move with the callout text...?

ErinBecker:
Looks good to me. I changed the follow up questions after the TPI challenge. Basically, instructors tend to have deep-seated doubts about the validity of educational research in general (and even more generally, people tend not to accept feedback) and I think we should avoid language that encourages them to trust their own instincts about their instruction above what the research shows. 

@raynamharris - I'm not sure about the git-iquite about adding commits on top of someone else's branch. If this isn't the way I'm supposed to do it, could you please let me know and I'll avoid in future. :-)

raynamharris:
@lexnederbragt I'm not sure which callout indicators didn't get moved... Can you point me to the lines or correct them?

@ErinBecker As far as I'm aware, adding commits on top is a perfectly fine way to do things, but I'm also pretty new at this :)

lexnederbragt:
Sorry, I hadn't seen that the {: .challenge} now is supposed to be after the text, it used to be before.

PR #220, opened 2016-06-10
new challenge: concrete DC SWC learning objectives

raynamharris:
add a more specific "evaluate learning objectives" challenge

### DIFF
--- a/_episodes/19-lessons.md
+++ b/_episodes/19-lessons.md
@@ -174,6 +174,31 @@ is a good longer discussion of these issues.
 > * Understand data slicing/manipulation.
 {: .challenge}
 
+
+> ## Evaluate real SWC and DC Learning Objects
+>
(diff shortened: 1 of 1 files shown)


ErinBecker:
I like this a lot. I'm a little worried people will get distracted by reading a bunch of learning objectives and spend time on that rather than doing the exercise. Can we add to the instructor's notes that instructor should either pick a specific lesson or give everyone a really short time limit (<30 seconds) to choose an objective?

raynamharris:
Thanks @ErinBecker. I would prefer the second instructor note, suggesting a time limit rather than limiting it to a specific lesson. A specifically identified learning objectives that covered the breadth of programming skills and objectives that spawned levels of taxonomy. 

In our class, learners chose the python, git, unix, and R ggplot learning objects but not the teaching or ecology objective, and they did so pretty quickly.
See lines 577 - 629 in our [class etherpad](http://pad.software-carpentry.org/2016-06-08-ttt-online)

Issue #231, opened 2016-06-13
Add intro episode


raynamharris:
- current intro lesson to be renamed something like "teaching philosophies" and moved to later in the day
- new intro episode will contain: welcome message, overview of training, and a short multiple choice questions

Goals:
- the multiple choice question will assess about prior knowledge of SWC/DC and some other things, which will help guide the instructors
- the overview should help get all the students on the same page, so they know where they are and where they are going for the whole 3 day training session

raynamharris:
Okay, I created my first pass at a welcome episode here: Create welcome episode 00-welcome.md #232.  Feedback on style and content is welcome.

gvwilson:
Closed by #232 - thanks!

PR #232, opened 2016-06-13
Create welcome episode 00-welcome.md 

raynamharris:
First attempt to address [issue #231](https://github.com/swcarpentry/instructor-training/issues/231) on adding a welcome episode. 

These are the first three multiple questions that come to mind. A variation of the 2nd is part of the post-workshop questionnaire. I'm open to adding more or rewording these or any other suggestions. 

Also, I call this lesson 00, but it could be 01, especially if the current episode one is moved to later.

### DIFF
--- a/_episodes/00-welcome.md
+++ b/_episodes/00-welcome.md
@@ -22,15 +22,15 @@ keypoints:
 
 To begin your class, the instructors should give a brief introduction that will convey the instructors': 1) capacity to teach the material, 2) accessibility/approachability, 3) desire for student success, and 4) enthusiasm. Tailor your introduction to the students’ skill level so that you convey competence (without seeming too advanced) and demonstrate that you can relate to the students. Throughout the workshop, continually demonstrate that you are interested in student progress and that you are enthusiastic about the topics.
 
-Students should also introduce themselves. At the very least, everyone should add their name to the Etherpad, but its also good for everyone at a given site to know who all is done in the group. Note: this can be done while setting up before the 9 am start of the class. 
+Students should also introduce themselves (preferably verbally). At the very least, everyone should add their name to the Etherpad, but its also good for everyone at a given site to know who all is in the group. Note: this can be done while setting up before the start of the class. 
 
 ## Workshop Overview
 
 This two-day workshop has two goals:
-- Introduce participants to a handful of key research findings and show how they can be used to help people learn better and faster.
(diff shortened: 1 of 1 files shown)


ErinBecker:
I like this and think it gives a good (short) contextual background for the whole 2 day event. I specifically like how you used a question about making MCQ questions - good lead in for first section of content.

ErinBecker:
Playing around with the idea of changing the answer options for the three mcqs so that they match our guidelines for good mcq questions. :-) Maybe not necessary.

raynamharris:
I think the question should definitely match our guidelines for good mcq questions :)

ErinBecker:
Or we can use it after the fact as an example of what not to do and ask them to improve it for the exercise? I know we already have them writing an MCQ question, improving one would be faster than writing from scratch, but wouldn't give them the prep for the teaching demo it's currently tied to.

ErinBecker:
@raynamharris What do you think of adding an MCQ asking about their previous experience teaching? This could either take the total up to 4 MCQs or replace the one about live coding. Would it be useful for us to know if the trainees are coming in with previous teaching experience and/or highly ingrained ideas about teaching?

raynamharris:
@ErinBecker let's go ahead an add one. Once all four are there, we can decide if one question should be cut or it the live coding question can be interviewed into the general previous teaching question

ErinBecker:
@raynamharris - I added a question, but am a little worried that the order of options will imply a hierarchy with respect to university vs. K12 teaching. Any ideas?

raynamharris:
@ErinBecker If the distinction between K-12 or university is not important, you could merge A and B to say something like "The lead instructor for a year- or semester-long course", which would distinguish them from being a teaching assistant or teaching short workshops. I do sorta see the hierarchy there, but it doesn't bother me. You could totally mix it up (say C, E, D, A, B)

ErinBecker:
@raynamharris I think it's important for us to know whether they have K12 teaching or university teaching experience. If they've taught K12 they've had formal pedagogical training and likely practice-based training, so are more likely to be on board with the curriculum. We could potential use folks like this as resources during the training (e.g. pair them up strategically). 

When I was doing graduate TA training, there were two TAs in the group who had taught K12. They were fantastic resources and very useful for helping get others "bought-in" to the idea of practice-based teaching.

gvwilson:
I like where this is going - please merge as soon as you're both happy with wording, and thank you.

ErinBecker:
I'm happy with it!

Issue #229, opened 2016-06-13
Remove exercises we aren't using


ChristinaLK:
Remove stuff.  In particular, remove exercises we aren't using, trim text.

ChristinaLK:
I didn't get to this this week and am trying to go off the grid this weekend, so will be tackling this first thing Monday morning.

ChristinaLK:
I didn't get to this this week and am trying to go off the grid this weekend, so will be tackling this first thing Monday morning.

gvwilson:
Can this now be

PR #236, opened 2016-06-20
Removing unused exercises

ChristinaLK:
Addresses #229 

This is completely based on what *I* do - if I've removed an exercise that you have consistently used, please comment and I'll put it back in. 

I'm also converting "challenges" that are really discussions to the "discussion" style.


raynamharris:
Thanks @ChristinaLK !  I agree almost completely!

In episodes/18-motivation.md, I like to say on of the things you've suggested deleting: 
"We all have a different motivation for teaching, and that is a really good thing! SWC wants instructors with diverse backgrounds because you each bring something unique to our community. Read the [testimonials]({{ site.github.url }}/guide/#testimonials) from other instructors, to see what motivates them to teach. **After this course*** Write a short explanation of what motivates you to teach. Save this as part of your teaching philosophy for future reference."

I agree that this shouldn't be a challenge, but I would like to keep the prose in the lesson in some form or other.

ChristinaLK:
I agree!  I actually incorporated it into a callout above, but I like the idea of separating the exercise and making it "for later."

ChristinaLK:
Wait to merge until #238 is addressed.

ChristinaLK:
If there are no other comments, this should be ready to merge.

PR #241, opened 2016-06-21
changing testimonial to be less wordy

ChristinaLK:
I think this pretty much sums it up.


_extras/guide.md
```diff
@@ -538,23 +538,7 @@ scheduled.
...
-> tell me that I'm not here to live for myself,
-> but that we're all in this together.
->
-> Most of all, I love sharing ideas with people.
-> The best moments in life, for me,
-> are facilitating intellectual insights along with personal connections.
-> Those moments are what build a community that continues living in hope and adventure,
-> especially important in times of fear and division.
-> I teach to make those moments possible for myself and for others.
+> <img src="/fig/guide/christinalk_whyiteach.png" width="50%">
```
gvwilson: Happy to have both, but the image alone fails our accessibility guidelines :-)

ChristinaLK: I had alt text in my original commit - oops.  

What would be necessary? alt text + title?

gvwilson: Alt text and a title would be fine :-)

PR #242, opened 2016-06-22
change Home to icon, move License to footer and squash searchbar

tomliversidge:
This will 

- change the text "Home" to a home icon
- move "License" menu item to footer
- shrink the width of the searchbar

### DIFF
--- a/_includes/footer.html
+++ b/_includes/footer.html
@@ -19,11 +19,13 @@
     </div>
     <div class="col-md-6" align="right">
       <h4>
-	<a href="{{ repo_url }}">Source</a>
-	/
-	<a href="{{ repo_url }}/blob/gh-pages/CONTRIBUTING.md">Contributing</a>
-	/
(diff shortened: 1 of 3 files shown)


tomliversidge:
capture2

tomliversidge:
capture

gvwilson:
Thanks for the PR - happy to shrink the width of the searchbar, might be 
allowed to change "Home" to an icon, but "License" will have to stay in 
the top menu (our lessons being CC licensed is a big part of our pitch).

Cheers,

Greg

tomliversidge:
Ok. Can any of the non- drop down menu items either go in an existing drop down or be grouped into a new one? I.e could License and Code of conduct be grouped together into a new drop down?

Or can everything be made smaller? And with less gaps between...

Trying to think of ways to save space!

gvwilson:
Sticking with existing layout.

PR #245, opened 2016-06-28
Moving overview of existing lesson materials to Carpentries lesson.

gvwilson:
Based on discussion with @ErinBecker.


ChristinaLK:
can this wait to be merged until after the Oslo training?

ErinBecker:
@ChristinaLK I think the plan was to do this after Oslo. @gvwilson Can you confirm?

gvwilson:
Yup - might merge between Oslo and SciPy, but that depends on thumbs-up.

ErinBecker:
This reorganization looks good to me. I think having the description of content when we introduce the organizations will be good.

ChristinaLK:
See another take on this in #256

PR #250, opened 2016-06-29
Move watching the live coding videos before top 10

lexnederbragt:
I suggest to first watch the two videos, collect and discuss feedback on them, and only *then* go over the top 10. Hopefully most of the items will have been treated by then.

### DIFF
--- a/_episodes/13-live.md
+++ b/_episodes/13-live.md
@@ -74,6 +74,13 @@ another.
 > from both a learner's and an instructor's point of view?
 {: .challenge}
 
+> ## The Bad and the Good
+>
+> Watch this video of [live coding done poorly][live-coding-bad]
+> and this video of [live coding done right][live-coding-good]
+> as a group and then summarize your feedback on both.
+{: .challenge}
+
 ## Live Coding Top 10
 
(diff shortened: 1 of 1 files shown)


ChristinaLK:
I would also suggest moving these *before* the discussion of live coding advantages/challenges - then people have more of a context for answering these questions.

lexnederbragt:
>I would also suggest moving these before the discussion of live coding advantages/challenges - then people have more of a context for answering these questions.

I don't agree. The discussion on (dis)advantages for instructors and learners sets the stage very well, and generates the necessary context for interpreting the videos.

ChristinaLK:
Having seen it in person, it worked the way you have it now.  
I think I just want to make sure that people know what live coding *is* so they can come up with realistic advantages/challenges.  But if you define live coding at the start and make sure that's clear to everyone, then you could go straight into advantages/disadvantages.

ChristinaLK:
I'd like to merge this - @gvwilson, can I steal it and do so?

gvwilson:
@ChristinaLK please go ahead and merge.

PR #251, opened 2016-06-30
Improve description of Peer Instruction

lexnederbragt:
Having students see the results of the first vote helps direct the discussion afterwards.


ErinBecker:
There's actually a pretty extensive debate about whether (and when) instructors should show the polling results. It ends up depending on what fraction of the students got the answer right in the first round. Since we're not trying to teach them the mechanics of PI to use in our workshops, I don't think we should take the time to address this controversy. I recommend leaving the discussion as is.

ChristinaLK:
I agree with Erin - I think there's not a clear benefit to showing the results to the students.

lexnederbragt:
Ok, you convinced me, thanks! Closing.

PR #254, opened 2016-06-30
Clarifying 2x2 plot (grid)

lexnederbragt:
what-to-teach.png is not a 2x2 plot. This adds a bit of clarification to the exercise. The alternative is to redraw the plot as a 2x2 grid.


raynamharris:
Thanks!

PR #257, opened 2016-06-30
intro material - also to discuss

ChristinaLK:
More surgery.  :P 

* Went straight from welcome -> mental models/formative assessment/mcq -> terms before the break
* shortened terms section
* changed MCQ quiz to 2 questions, where the second question attempts to have answers with diagnostic power.  ;)


_episodes/01-welcome.md
```diff
@@ -20,111 +20,113 @@ keypoints:
...
-> 5.  Strongly disagree   
->
-> *If you teach, how often do you incorporate live coding in your classrooms? If you don't teach, how often have your instructors used live coding?*
->
-> 1.   Frequently
-> 2.   Occasionally
-> 3.   Rarely
-> 4.   Never
+> 
+> *Which of these questions assesses flaws in a student's mental model of a domain?*
```
ErinBecker: I like this!

gvwilson:
@ChristinaLK @ErinBecker if you both still like this, I'll merge it this week.

ChristinaLK:
I'd like some more comments - can @raynamharris @lexnederbragt @karinlag take a look?

ErinBecker:
I'm +1 to merging this, but agree would be great if we got more input from @raynamharris @lexnederbragt @karinlag.

raynamharris:
I like most of the changes.  The only think I worry about with MCQ is the lack of an answer key... I don't use python, and the unix cut command is something I always have to Google, so I would have to remember to do some research on those before teaching...

_episodes/01-welcome.md
```diff
@@ -20,111 +20,113 @@ keypoints:
...
-Now that we have a better idea of everyone's prior knowledge and
-familiarity with some of the SWC and DC teaching practices, we're
-ready to begin our training. Our goal is that by the end, you will
-have acquired some knew knowledge, confidence, and skills that you can
-use in your teaching practice in general and in teaching SWC and DC
-workshops specifically.
-
-Let's dive into the theory and practice behind designing multiple
+Now that we have a better idea of everyone's prior knowledge, 
+let's dive into the training.  Our first topic will be
```
lexnederbragt: --> Our first topic will be Cognitive Development and Mental Models

(or at least not just MCQs, even though these are an element of it)

_episodes/03-terms.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: "Introduction"
```
lexnederbragt: Should no longer be called 'Introduction', but something like 'important terms'?

_episodes/03-terms.md
```diff
@@ -17,65 +17,6 @@ keypoints:
 - "Pedagogical content knowledge: connects general understanding of teaching to domain-specific content."
 ---
 
-This section outlines the structure of this two-day course and provides some
-essential terms for discussion.
-
-## Practice and Feedback
```
lexnederbragt: Could you move the 'Practice and Feedback' part to the beginning of 05-performance?

lexnederbragt:
I added some comments, but otherwise I'm +1. I see @raynamharris' point on the unix cut command not necessarily being know by all instructors. But the same holds for the python question. I propose to leave them as as and think about more domain/language agnostic alternatives that any instructor in training can relate to after the merge.

raynamharris:
Thanks @lexnederbragt and @ChristinaLK  +1

karinlag:
Looks good to me too!

ChristinaLK:
Okay @gvwilson, all comments addressed and now up to date w/ the latest version.  Can we merge this, and @raynamharris and I will test in production when we teach at the end of the month?

Issue #261, opened 2016-07-05
Use table rather than list of existing lesson material


lexnederbragt:
At http://software-carpentry.org/lessons/ there is a table listing lessons, with icons that link to the website, repo and reference for each lesson. It would be nice to have a similar table in [`19-lessons.md`](http://swcarpentry.github.io/instructor-training/19-lessons/), but then with icons that link to the website, repo and _instructor guide_ (for those repos that have one).

ErinBecker:
@lexnederbragt I'm inclined to suggest that we simply link out to the lessons page for SWC and DC rather than reproducing those tables here. If there are changes to the curriculum (either additions or deletions), we would need to update in two places, which can only cause problems down the road. :-)

I'm happy to make this change if no objections.

lexnederbragt:
I agree, but while thinking about this I realised the instructor training lesson material may benefit from having a section on 'how to prepare for teaching a lesson' (perhaps in 'Carpentry Teaching Practices'?)

karinlag:
Oooh, @lexnederbragt I really like this one. Maybe we should send this one out on the email list, so that we can get a sense of what people do? I am pretty sure that what I do might not be what everybody else does...

raynamharris:
I agree with @lexnederbragt. There is already a table (albiet with less information) about SWC lessons, but the DC lessons are in table form. At the very least, the DC lessons should be put in a table for consistency.

ErinBecker:
Closed via https://github.com/swcarpentry/instructor-training/issues/406

PR #262, opened 2016-07-07
Made list of SWC lessons into a table

lexnederbragt:
Addresses #261. However, for Software Carpentry only, as the Data Carpentry lesson links actually refer  to multiple repos each with their own github repo and instructor guide.

No fancy icons, though, not sure how to get them added. As pngs?


gvwilson:
Thanks.

PR #268, opened 2016-07-16
Fixes and typos

andreabedini:
Here are some trivial fixes.

* Add *.pyc to `.gitignore`
* Fix typo `{.callout}` -> `{: .callout}`
* Add missing future statement in `bin/util.py`
* Remove some unnecessary code in `bin/lesson_check.py`

I apologise for the single PR but they are fairly trivial and can be cherry-picked independently.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 .sass-cache
 _site
 __pycache__
+*.pyc
--- a/_episodes/19-motivation.md
+++ b/_episodes/19-motivation.md
@@ -159,7 +159,7 @@ to make sure they're doing at least a few of these things.
 > learners respond to an instructor's enthusiasm,
 > and instructors need to care about a topic in order to keep teaching it,
 > particularly when they are volunteers.
-{.callout}
+{: .callout}
 
 > ## Why Do You Teach?
 >
--- a/bin/lesson_check.py
+++ b/bin/lesson_check.py
@@ -388,10 +388,6 @@ class CheckBase(object):
 class CheckNonJekyll(CheckBase):
     """Check a file that isn't translated by Jekyll."""
 
-    def __init__(self, args, filename, metadata, metadata_len, text, lines, doc):
-        super(CheckNonJekyll, self).__init__(args, filename, metadata, metadata_len, text, lines, doc)
-
-
     def check_metadata(self):
         self.reporter.check(self.metadata is None,
                             self.filename,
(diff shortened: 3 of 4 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 .sass-cache
 _site
 __pycache__
+*.pyc
```
gvwilson: Good catch - I'll incorporate into the styles repo used for shared material.

_episodes/19-motivation.md
```diff
@@ -159,7 +159,7 @@ to make sure they're doing at least a few of these things.
 > learners respond to an instructor's enthusiasm,
 > and instructors need to care about a topic in order to keep teaching it,
 > particularly when they are volunteers.
-{.callout}
+{: .callout}
```
gvwilson: Already fixed on the incoming PR for the 2016-06 release.

bin/lesson_check.py
```diff
@@ -388,10 +388,6 @@ def get_loc(self, node):
 class CheckNonJekyll(CheckBase):
     """Check a file that isn't translated by Jekyll."""
 
-    def __init__(self, args, filename, metadata, metadata_len, text, lines, doc):
-        super(CheckNonJekyll, self).__init__(args, filename, metadata, metadata_len, text, lines, doc)
-
-
```
gvwilson: I always include explicit constructors to make it clear to novices where to add new initialization code.

bin/util.py
```diff
@@ -1,3 +1,5 @@
+from __future__ import print_function
```
gvwilson: We're a pure-Python 3 shop now :-)

gvwilson:
Thanks Andrea - commented.

bin/util.py
andreabedini: Very happy with that! But then the scripts should have python3 in the shebang right?

bin/lesson_check.py
andreabedini: Ah ok, perhaps add a comment line about this?

.gitignore
andreabedini: Well, if you are a python3 only shop, you won't need this :)

bin/util.py
gvwilson: Ah - Anaconda does create an alias 'python3' for that Python.  Sure, we 
can do that.

PR #267, opened 2016-07-16
Fix missing and spurious references

andreabedini:
Some references were missing and some were spurious. I used the script below to check. Perhaps it could be included into `lesson_check.py`.

``` python
import re
from collections import defaultdict, namedtuple
from glob import glob

re_refs = re.compile('\]\[([a-z-]+)\]', flags=re.M)
re_defs = re.compile('^\[([a-z-]+)\]:\s*(.*)', flags=re.M)

fns = glob('_episodes/*.md')

def get_file_refs(fn):
    text = open(fn).read()
    return set(re_refs.findall(text))

def get_file_defs(fn):
    text = open(fn).read()
    return dict(re_defs.findall(text))

file_refs = {fn: get_file_refs(fn) for fn in fns}
file_defs = {fn: get_file_defs(fn) for fn in fns}

Suggestion = namedtuple('Suggestion', ['file', 'url'])
reverse_index = defaultdict(list)
for fn, refs in file_defs.items():
    if not refs: continue
    for key, url in refs.items():
        reverse_index[key].append(Suggestion(fn, url))

for fn in fns:
    defined_keys = set(file_defs[fn].keys())
    missing = file_refs[fn] - defined_keys
    spurious = defined_keys - file_refs[fn]

    if missing or spurious:
        print('\nFound issues with', fn)

    for ref in missing:
        print("Missing reference", ref)
        for suggestion in reverse_index[ref]:
            print('Suggestion for', ref, suggestion.url, 'as defined in', suggestion.file)

    if spurious:
        print("Spurious references", ', '.join(spurious))

```

PR #269, opened 2016-07-18
Fixed links in _extras/exercises.md that were lost.

evanwill:
noticed that links were lost in `_extras/exercises.md`. 
It looks like links in `_episodes/` were updated in #267  , but not on in this directory.


gvwilson:
Thanks!

PR #277, opened 2016-07-27
Fix broken links in Motivation and Demotivation lesson

karawoo:
A few of the links to other lessons were broken.


gvwilson:
Thanks @karawoo - I really need to add link checking to the tools.

PR #283, opened 2016-08-06
Correcting attribution

gvwilson:


lexnederbragt:
Ah, good to know. We could still reference a tweet, but then use this one instead? https://twitter.com/jduckles/status/684794574003245056. On the web, this includes the tweet quoting it.

PR #290, opened 2016-08-16
Checkout: add teaching demo lesson/segment examples

lexnederbragt:
Based on some feedback from a instructor trainee, there is a danger of confusion when trainees need to choose a lesson to teach from. For SWC all is well, what is a lesson and an episode is clear from the webpage. For DataCarpentry, however, the situation is a bit more blurry. http://www.datacarpentry.org/lessons/ lists workshops which each have lessons. But, for example, the R-genomics lesson http://www.datacarpentry.org/R-genomics/ has its episodes called `lesson 00`, `lesson 01` etc. Perhaps we need to ask DC to adjust that. 

At any rate, this PR adds an example lesson and segment that trainees may be asked to demonstrate for both SWC and DC.


gvwilson:
thanks Lex

PR #301, opened 2016-09-08
changes to instructor's notes

ChristinaLK:
folding in material that I cut from the introduction
also, rearranging so material most relevant to instructors is on top
and removing outdated discussion session description


lexnederbragt:
OK for me - but we should at some point (yeah, I know...) read through all pages, including this one, to check for consistency, redundancy and unnecessary lengthiness :-)

ChristinaLK:
I don't delete, I reorganize, haha.

PR #302, opened 2016-09-09
slight time changes

ChristinaLK:


lexnederbragt:
How does this affect overall length?

ChristinaLK:
puts lunch at noon on both days, finish at 16:10 on the first day, 16:00 on the second.

PR #303, opened 2016-09-16
Links to Key Reading .pdf are broken.

humberto-ortiz:
I think the urls are missing "instructor-training".


gvwilson:
Thanks for spotting this.

PR #305, opened 2016-09-20
Link fixes

lexnederbragt:
Many links in 19-lessons.md and 21-carpentries.md didn't render.
This is a brute force fix, but it could also be a jekyll issue.

### DIFF
--- a/_episodes/21-carpentries.md
+++ b/_episodes/21-carpentries.md
@@ -30,27 +30,27 @@ and on the final steps toward certification.
 
 ## History
 
-[Software Carpentry][swc] was co-founded in 1998 by Brent Gorda and
+[Software Carpentry]({{ site.swc_site }}) was co-founded in 1998 by Brent Gorda and
 Greg Wilson, who identified a need for best practices training in
 research computing.  After several iterations, the current model of
 two-day workshops with a standard curriculum emerged in 2010-11.
 After intermediate support from various organizations, it became an
 independent non-profit organization called the [Software Carpentry
-Foundation][scf] (SCF) in 2015.  The SCF is now responsible for all
+Foundation]({{ site.swc_site }}/scf/) (SCF) in 2015.  The SCF is now responsible for all
 aspects of Software Carpentry's operations.
 
 > ## History Lesson
 >
 > For more on Software Carpentry's history,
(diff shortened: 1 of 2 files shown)


arokem:
As far as I can tell, this is a jekyll issue resolving the liquid tags in locally defined links. 

It seems that this can be resolved by surrounding the links with angle brackets:

`[dc-blog]: {{ site.dc_site }}/blog/` => `[dc-blog]: <{{ site.dc_site }}/blog/>`

lexnederbragt:
Jekyll, OMG :-)

Your solution works locally (testing the site on my laptop), at least.

_episodes/21-carpentries.md
```diff
@@ -378,26 +378,26 @@ you are very welcome to put your name forward as a candidate.
 > What would you add, and what would you drop to make room?
 {: .challenge}
 
-[dc-blog]: {{ site.dc_site }}/blog/
+[dc-blog]: <{{ site.dc_site }}/blog/>
```
gvwilson: Hm - `<url>` is Markdown's way of saying "show `url` but also make it a link".  The footer links (like this one) are a way of saying "I want to use a shortened form in the main text, and put the full URL at the bottom", so putting this in angle brackets shouldn't work.  I'll dig...

arokem: It shouldn't... but it seems to anyway, without showing the URLs explicitly anywhere on the page, as far as I could tell. At least locally on my laptop (and sounds like on @lexnederbragt's as well). It doesn't work on your local setup?

gvwilson:
Nope - I'm running Jekyll 3.1.2 with Kramdown 1.11.1:

$ jekyll --version
jekyll 3.1.2
$ which kramdown
/usr/bin/kramdown
$ kramdown --version
1.11.1

lexnederbragt:
```
$ jekyll --version
jekyll 3.0.1
$ kramdown --version
1.9.0
```

gvwilson:
*sigh* I dunno...  The fix I pushed (which inlines the URLs using site 
variables) is what I've done elsewhere, so I vote to merge that and come 
back to this when someone has more time to dig into Jekyll templating.

gvwilson:
Merged #307 instead.

Issue #311, opened 2016-09-29
add timing for first video exercise


ChristinaLK:
It doesn't say how long it should be.  It sounds like it should be 3 min, but that ends up being a little long...
cc @gvwilson

raynamharris:
I like 2 minute videos.

PR #321, opened 2016-10-20
Add max length of first video

lexnederbragt:
Addressing #311, this is what we used as length last time I did a TTT.


ChristinaLK:
That's what I remember as well.  Can at least 1-2 more people weigh in?

PR #338, opened 2016-11-02
Message to send out when training sessions time out

gvwilson:


files/messages/expiry.txt
```diff
@@ -0,0 +1 @@
+It's been several months since you took part in the Software Carpentry and Data Carpentry instructor training class, and we are now wrapping up your cohort.  According to our records, you have not yet completed the final checkout exercises - if you have, and our records are inaccurate, please let me know, and if you haven't, but are still intending to do so, please get in touch as quickly as you can so that we can work out a plan.
```
lexnederbragt: Not everyone may get the word 'cohort'. Perhaps skip the ', and we are now wrapping up your cohort' part?

lexnederbragt: Split the second sentence in two:

"If you haven't completed the checkout exercises, but are still intending to do so, please **get in touch as quickly as you can** so that we can work out a plan.

ErinBecker:
I know there was discussion several months ago about having an explicit deadline for completing checkout. Was it ever decided what the deadline would be? If so, was it communicated to trainees at the most recent batch of TTT events? If so, I think the last phrase should be changed from the vague "as quickly as you can" to the actual deadline for that person according to the new policy.

gvwilson:
`_extras/checkout.md` now includes:

> We would like to get everyone through this process within two months 
> of training. Anyone who has not completed within 90 days should 
> contact us <mailto:%7B%7B%20site.email%20%7D%7D> to discuss reasons 
> and options.
and people are told during training about the 90-day limit; this message 
is what we send when that limit has been reached.

raynamharris:
I have a question about this. What do I do when students email and ask for an extension, with what I believe to be a good reason?

gvwilson:
If you believe it's a good reason, say 'yes' - I've tried to accommodate 
family issues, job changes, and thesis deadlines as much as possible.  
We don't have a way to mark extensions in AMY yet, but you can edit the 
Person record and add an administrative note.

raynamharris:
Will do

raynamharris:
Are we also sending out messages before the 90 days are up (maybe at T-45 or T-14 days)  that say like "your training checkout time will expire in xx days." 

I feel like this way we can give them a few prompts to sign up. And, it would be great if our etherpads accommodated discussion sessions and demos out say 45 days in the future...

gvwilson:
Haven't had the resources to do this in the past, but please file a 
feature request for AMY - if we can get notifications (somewhere), then 
maybe someone will have time to follow up on them.

raynamharris:
Thanks for the clarification @gvwilson.

I can fill our a feature request, but I'd like to see what other people's experience is with the 90 day limit we set a while back. Maybe we can figure out when is best to send out reminder prompts and the best way of doing it.  I've reopened the discussion on [an older issue](https://github.com/swcarpentry/instructor-training/issues/234)  where some stats on complete rates prior to June were posted.

PR #344, opened 2016-11-09
Message to send to passive 'participants' in discussion sessions

gvwilson:
@maneesha can you please review?


maneesha:
Looks good to me.

maneesha:
Actually, giving it more thought... perhaps we could include language asking them if anything blocked them from participating.

> If you experienced any difficulties partcipating, please let us know so we can explore ways to be of more support.  This could include the language the sessions are conducted in, the platform hosting the video calls, or other considerations.

karinlag:
I agree with @maneesha says above. We should first ask if they blocked on anything, and if not, then ask them to attend another one. More overhead, but still.

PR #347, opened 2016-11-14
Adds link to introductory socrative MCQ import

lexnederbragt:
The google form is not accessible for viewing the results nor for modifying it


raynamharris:
Cool!  Should we make copies of the form before administering it so that the responses are always from that day's workshop?

lexnederbragt:
If you import the quiz into your Socrative account, there are no answers with it. Then any answers you get will be from that day's workshop. For the next workshop, duplicate the quiz again inside your own Socrative account.

If SWC/DC 'officially' endorse Socrative, there should perhaps be made a page to how to use it for workshops.

ErinBecker:
Merged! Thanks @lexnederbragt and @raynamharris.

PR #348, opened 2016-11-17
Correct link formatting and syntax

ethan-nelson:
In the _lessons_ episode, a link to one of the papers has a typo in the formatting.

I also changed the base url variable so it would work correctly for instances not hosted through GitHub Pages (the `{{page.root}}` syntax is used elsewhere in the repository).


gvwilson:
Thanks for catching this.

PR #351, opened 2016-12-02
second parsons problem challenge

raynamharris:
Given that the beginning of this lesson talks about "Extraneous load" I'm not really a fan of having the parson's problem challenge have an unnecessary line of code in it. So, I added a challenge question that is just about jumbling and un jumbling a problem.

### DIFF
--- a/_episodes/11-load.md
+++ b/_episodes/11-load.md
@@ -184,11 +184,19 @@ and so on.
 
 > ## Parsons Problems
 >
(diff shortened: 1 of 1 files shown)


kcranston:
I like this second exercise even better than the first (and it gets people looking at the lessons!). I'd merge, but I don't yet have write access to this repo.

gvwilson:
You should have write permission now.
- G

PR #354, opened 2016-12-04
Updating readings and making use of Lee's paper on inclusivity.

### LINKED ISSUE #298: Add reference to Cynthia Lee's paper on creating inclusive community
And make it part of the required upfront reading.

gvwilson:
1.  Added reference to Lee's paper on making classrooms inclusive to pre-workshop reading list.
2.  Added descriptions of recent books and papers to reference material.

Closes #298.


smcclatchy:
Greg,
I used a couple of the suggestions in this paper when I started my computational biology course this term. I have it taped to the wall now for frequent referral, since it contains a lot suggestions and a few warnings. It's a great addition.

ErinBecker:
I think there's a lot of great advice in this article and think it would be a great pre-reading for a seminar on teaching CS university courses (or other STEM courses at most grade levels). However, I think here as always we need to be conscious of our audience and our mission. 

Is our primary objective to educate people on how to be effective instructors for our context and content, or good instructors in general? If the first, I note that most of the suggestions in the new article are not relevant to our workshops (as they assume recurring contact over the course of a term, ongoing homework assignments, use of slide decks, etc). 

If the later, I note that most of our instructors are not going to be teaching CS classes - and many will not be teaching courses on that side of the spectrum. This is particularly true of people taking the training to become Data Carpentry instructors (and in future we also have to think about Library Carpentry instructors, etc). What does it say to a biology graduate student, or a staff librarian when this is one of the assigned readings the day before the course? I think it says that the class is for people who are planning on teaching CS. I don't speak for everyone, of course, but I know that this would have shut me off and maybe even made me question if this was the right training for me. 

On a broader note, I think that we should only assign pre-class readings if we are planning on using the content from those readings in class. Any assignment that doesn't directly relate to course content comes across as busy work and create negative attitudes in the learners. 

So for these reasons, and not because I disagree with anything in the article or think it's not a great set of recommendations, I would vote strongly against making this a required pre-class reading.

raynamharris:
The comments by @smcclatchy and @ErinBecker are both very good. 

Is there a way we can add this to a "suggested readings for after you complete instructor training" rather than making it a required reading before training @gvwilson?

gvwilson:
Thank you all for your comments - I have moved the shopping list of motivational techniques into the reference material (since rattling off bullet points doesn't have much effect), and used the space freed up to add a couple of paragraphs referring to Lee's paper on inclusivity, and a "remind your future self" exercise that won't take up much class time.  Comments would be very welcome.

files/messages/welcome.txt
```diff
@@ -1,9 +1,10 @@
 Hello everyone, and thank you for signing up for our upcoming instructor training session. In order to prepare:
 
-1.  Please read these two short papers, which provide a brief overview of some key evidence-based results in teaching:
+1.  Please read these three short papers, which provide a brief overview of some key evidence-based results in teaching:
 
     * "The Science of Learning" (https://swcarpentry.github.io/instructor-training/files/papers/science-of-learning-2015.pdf)
     * "Success in Introductory Programming: What Works?" (https://swcarpentry.github.io/instructor-training/files/papers/porter-what-works-2013.pdf)
+    * "What Can I Do Today to Create a More Inclusive Community in CS?" (https://swcarpentry.github.io/instructor-training/files/papers/lee-create-inclusive-community-2015.pdf)
```
lexnederbragt: This does not really "provide a brief overview of some key evidence-based results in teaching"...

lexnederbragt:
I am also not in favour of adding the Lee text as required pre-workshop reading  - as much as I love it. I think there is room for a third short text, I just don't know what it would be.

PR #361, opened 2016-12-13
Fix links to workshop template website and repo

ngoldbaum:
Right now these links are broken on the [rendered page](http://swcarpentry.github.io/instructor-training/21-carpentries/). This is the first time I've ever messed with Jekyll so it's entirely possible this is wrong.


gvwilson:
Thanks for the PR - using `site.workshop_repo` in both locations.

PR #363, opened 2016-12-14
Update passive-discussion.txt

maxim-belkin:
Reworded the message for passive participants to not discourage them from further participation


lexnederbragt:
See also #355 which addresses the same file. I like your changes so I took some if them into that PR - let me know what you think.

maxim-belkin:
I think one PR (#355) for a file is #goodenough :)

PR #365, opened 2016-12-16
Update checkout.txt

raynamharris:
I updated this post workshop email by added a link to the post-workshop survey. @davis68 @gvwilson 

I usually email @tracykteal @maneesha @ErinBecker first to get the workshop added to the survey. We should confirm that this is still the correct survey to use as a follow up.

### DIFF
--- a/files/messages/checkout.txt
+++ b/files/messages/checkout.txt
@@ -1,6 +1,10 @@
 Hi everyone,
 
-Thanks again for taking part in instructor training this week. In order to wrap up, we would like you to do the steps outlined in http://swcarpentry.github.io/instructor-training/checkout/:
+Thanks again for taking part in instructor training this week. I really enjoyed spending time with you, and I look forward to your continued involvement with the Software and Data Carpentry Communities. 
+
+We'd be very grateful if you could take a couple of minutes to fill in this survey (https://www.surveymonkey.com/r/post-instructor-training) to provide annoymous feeback about the workshop.
+
+In order to wrap up, we would like you to do the steps outlined in http://swcarpentry.github.io/instructor-training/checkout/:
 
 1. Submit something for one of our lessons.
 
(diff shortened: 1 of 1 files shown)


tracykteal:
Thanks for adding this. I'll check on the survey to make sure it's the right one. And we should run it by Kari too. We likely can do that when we meet in January. 

I think @maneesha can be the one to add the workshop link. It can go hand in hand with getting the instructor training event into AMY, but we'll need to talk about the point of contact there too.

raynamharris:
Sounds like a plan.  Thanks @tracykteal

files/messages/checkout.txt
```diff
@@ -1,13 +1,17 @@
...
+In order to wrap up, we would like you to do the steps outlined in http://swcarpentry.github.io/instructor-training/checkout/:
 
 1. Submit something for one of our lessons.
 
 2. Take part in an online discussion session.
 
 3. Teach a 5-minute demonstration lesson online.
 
-If you have any questions or suggestions about these steps, please contact your trainer or mail us at training-help@software-carpentry.org.  Please note that we would like everyone to complete all three steps within 90 days of the training course.  If you are not able to do this, please contact your trainer to discuss reasons and options.
+If you have any questions or suggestions about these steps, please contact me or mail all the trainers at training-help@software-carpentry.org.  Please note that we would like everyone to complete all three steps within 90 days of the training course.  If you are not able to do this, please contact me to discuss reasons and options.
```
maxim-belkin: I'd suggest removing "mail" so that it reads
> please contact me or all the trainers at ....

Another one, you can pre-populate subject line like so:
> `[training-help@software-carpentry.org](mailto:training-help@software-carpentry.org?Subject=Instructor%20training%20checkout)`

I had to use backticks to prevent GitHub's engine from rendering the link

raynamharris: I like the pre-populated suggestion!

tracykteal:
I checked in surveymonkey and confirm that it's the right link.

files/messages/checkout.txt
```diff
@@ -1,13 +1,17 @@
...
+In order to wrap up, we would like you to do the steps outlined in http://swcarpentry.github.io/instructor-training/checkout/:
 
 1. Submit something for one of our lessons.
 
 2. Take part in an online discussion session.
 
 3. Teach a 5-minute demonstration lesson online.
 
-If you have any questions or suggestions about these steps, please contact your trainer or mail us at training-help@software-carpentry.org.  Please note that we would like everyone to complete all three steps within 90 days of the training course.  If you are not able to do this, please contact your trainer to discuss reasons and options.
+If you have any questions or suggestions about these steps, please contact me or all the trainers at [training-help@software-carpentry.org](mailto:training-help@software-carpentry.org?Subject=Instructor%20training%20checkout). Please note that we would like everyone to complete all three steps within 90 days of the training course.  If you are not able to do this, please contact me to discuss reasons and options.
```
gvwilson: This is usually sent out as plain text in an email message, rather than as formatted Markdown, so URL-formatting like this won't work.

PR #366, opened 2016-12-18
Filling in steps for trainers on AMY

gvwilson:


lexnederbragt:
Sorry to be late, but the wording of

>  Oversee final demonstrations and mark them as complete in AMY.

contrasts with

> Note that trainers do not examine their own trainees

lexnederbragt:
Have a look at #367

raynamharris:
I don't think they conflict @lexnederbragt  We've made it clear in a few places that trainers don't examine their own trainees. This file just documents the steps for adding all pieces to AMY related to the trainining workshop, homework, demo session, and discussion. It doesn't mean one trainer is doing all four for one learner; rather, its just all the instructors in one piece that you consult as needed.  Correct @gvwilson ?

karinlag:
@raynamharris , that is how I read this at least.

PR #367, opened 2016-12-19
Move demo marking

lexnederbragt:
Trainers should not examine their own trainees


ErinBecker:
@lexnederbragt see https://github.com/swcarpentry/instructor-training/commit/5ed967d635bd3865df5d741b3323f3674c3dc598

PR #383, opened 2017-02-22
Define the VPS and SSH acronyms in the Carpentry Teaching Practices document

gawbul:
During our instructor training session (https://allegravia.github.io/2017-02-21-online-instructor-training/) a problem was highlighted in that some people didn't know the definition of the acronyms VPS and SSH. Perhaps links to these would be relevant?

### DIFF
--- a/_episodes/15-practices.md
+++ b/_episodes/15-practices.md
@@ -137,10 +137,11 @@ Learners use their own machines.
     two different sets of keyboard shortcuts for things like copying and
     pasting.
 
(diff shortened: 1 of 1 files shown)


mkuzak:
@gawbul thanks I think it's useful

raynamharris:
I agree. Its always good to have acronyms spelled out! Thanks!

ErinBecker:
Thanks @gawbul!

Issue #332, opened 2016-10-26
Trainers should not certify people from their own institution


gvwilson:
In order to avoid conflict of interest, trainers should not certify people working at their home institutions.

lexnederbragt:
I propose to do this as part of #328

gvwilson:
:+1:

ErinBecker:
@lexnederbragt Could you confirm whether this has been added to the documentation?

ErinBecker:
Closed due to redundancy with #384

PR #384, opened 2017-03-01
Clarify not to certify certain trainees

lexnederbragt:
Trainers should not certify people from their own institution or from the workshops they taught. Addresses #332 and more.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -445,6 +445,8 @@ Checklist for instructor trainers hosting a live-coding demo session as part of 
 ### Before the Demo
 
 *   Sign up to lead the demo at <http://pad.software-carpentry.org/teaching-demos>.
+*   Trainers should not certify people from their own institution or from the workshops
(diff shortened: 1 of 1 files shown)


ErinBecker:
I agree with not certifying people from their own institution, but why not from workshops they taught? I don't see any particular conflict of interest there. Do you? Has there been conversation around this somewhere I'm not aware? 

This requires Trainers to look at who is coming to their demo and message people ahead of time that they can't attend their demo if they happen to be from a training you taught. Nevermind that this requires Trainers to remember the names of everyone who was in their training (which I certainly don't!)

ErinBecker:
I see it on the Etherpad now:

"Please don't sign up for a teaching demo with a trainer who belongs to your institution.
Please don't sign up for a teaching demo with the trainer that trained you."

So if this is a policy that's already in place, I'm happy to adopt and adapt myself. :-)

PR #389, opened 2017-03-03
Update guide.md

raynamharris:
Moved sending the welcome email from a responsibility of the trainer to a responsibility of the coordinator. Is that the new plan or just a one off @maneesha @ErinBecker ?

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -23,6 +23,10 @@ for a look at how using checklists can save lives (and make many other things be
 >     *   If it is in person, book travel.
 > 1.  Get names and email addresses of attendees from host(s).
 >     *   Register those people in AMY.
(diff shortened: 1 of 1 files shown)


ChristinaLK:
-1
I think this email should come from the trainer -- this is your first introduction and way to connect with them before the training.

maneesha:
In the past couple of months since I've picked up on managing training events, I've found it challenging to get names of trainees from host sites.  What I've been doing for now is that I send a general email introducing everyone, then the trainer(s) can send a follow up introducing themselves and sharing any other info.

As we get better systems in place for sites to sign up for trainings and get a list of trainees to us, this may change.

ErinBecker:
Merged, thanks @maneesha, @raynamharris and @ChristinaLK.

PR #390, opened 2017-03-03
Update guide.md

raynamharris:
Added and rearranged some information to the what to do before a workshop section.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -27,13 +27,14 @@ for a look at how using checklists can save lives (and make many other things be
 
 > ## Setting Up [Trainer]
 >
+> 1.  Create an Etherpad (http://pad.software-carpentry.org/<nameofevent>
 > 1.  Set up a one-page website for the workshop using <https://github.com/swcarpentry/training-template> as a starting point.
 > 1.  Send the URL to the admins.
-> 1.  Check whether any attendees have special needs.
 > 1.  If it is online:
 >     *   Test the video conference link.
+>     *   Set up meeting with the hosts to make sure the bluejeans channel works and give you a change to meet "face-to-face"
+> 1.  Check whether any attendees have special needs.
 > 1.  Make sure attendees will all have network access.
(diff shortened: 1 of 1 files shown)


ChristinaLK:
etherpad changes look good.  
why remove checking for special needs?

raynamharris:
Yikes. I didn't mean to remove checking for special needs.

ChristinaLK:
I figured.  :)  This is why we review!

_extras/guide.md
```diff
@@ -27,13 +27,15 @@ for a look at how using checklists can save lives (and make many other things be
 
 > ## Setting Up [Trainer]
 >
+> 1.  Create an Etherpad (http://pad.software-carpentry.org/<nameofevent>
```
lexnederbragt: There have been experiments with Google Docs instead, perhaps mention that as well?

raynamharris: Yes, there have been. I think the important point to make would be that the link to the google doc should still be pasted in the "etherpad" section of the workshop template. 

Have you ever used google docs? Want to take a stab at adding those instructions?

lexnederbragt: I haven't used a google docs for a workshop (yet), but @gvwilson has...

mkuzak: I have used the doc. I agree that the link should go in place of etherpad link. One issue we had with the doc is, people cannot chat if they're not logged in in google account. This is bigger of an issue in online training, since it depends on chat a lot.

ChristinaLK:
Do we want to merge as is + open an issue to add the google doc instructions separately? @ErinBecker

ErinBecker:
@raynamharris Merged! 

@lexnederbragt  Could you add an issue for the google doc instructions?

lexnederbragt:
@ErinBecker Looks like #405 covers the Google docs issue

ErinBecker:
@lexnederbragt Great, thanks.

PR #393, opened 2017-03-03
Update styles to v9.1.1

rgaiacs:
New version of styles: https://github.com/swcarpentry/lesson-example/releases/tag/v9.1.1

Preview: https://rgaiacs.github.io/swc-instructor-training


ChristinaLK:
This looks good (as far as I can tell -- it's hard to review these kinds of "plumbing" changes).  

@tracykteal @maneesha do we have any upcoming workshops, i.e. would it be bad if this temporarily broke something?

PR #397, opened 2017-03-07
added google form version of red/green stickies

raynamharris:
@ChristinaLK I'm officially suggesting that we add this to the curriculums. I'm so glad you made the original one :)

### DIFF
--- a/_episodes/06-summarize.md
+++ b/_episodes/06-summarize.md
@@ -29,4 +29,8 @@ are still unanswered.
 > on the red.  Do *not* put your name on the notes: this is meant to
 > be anonymous feedback.  Add your notes to the pile by the door as
 > you leave to get coffee.
+>
+> For online workshops, you can use a google form to collect annonymous feedback. 
+> Make a copy of this one and change the name to that of your workshop. 
+> https://goo.gl/forms/AweE07dXtib6QDXE2
 {: .challenge}


ChristinaLK:
I forgot about this!  Glad it's been useful.  :)

I think this change is a good one -- will leave it open for more comments but will check in and merge later this week if no one else does.

lexnederbragt:
+1. But:

>Make a copy of this one and change the name to that of your workshop

 I couldn't make a copy. Can you try to enable editing of the form? I think that solves it.

raynamharris:
@lexnederbragt I've enabled editing, so you should be able to use it. 

The data collected in my most recent workshop was incredibly useful!!! They students had very positive reactions to our openness to critique and ability to change our behavior to accomodate suggestions for improvement.

lexnederbragt:
When I go to the form I can't see an edit link, but maybe that is also not the way to do it. Is there a link that according to google allows others to edit?

Not sure what is an easy way to achieve what we want - I'm still hoping for a system where you write questions in markdown and get them rendered as a fully working questionnaire online. One alternative is to sharing through Socrative for this form as well...

raynamharris:
Hi @lexnederbragt. No, you don't want to edit the form... you want to make a copy. You click the three dots up at the top and select make a copy. This will create your own and give you immediate access to all the responses, which you can share with co-teachers and other instructors/staff by adding collaborators.  
screen shot 2017-03-14 at 9 18 46 pm

I'm not opposed to Socrative, I just never use it.

lexnederbragt:
Here is a screenshot of what the form looks like in my browser(s). No three dots... :-(
And I am logged into google, so it's not that.

image

ChristinaLK:
I think you have to actually give people edit rights for them to copy the form.  :P  Would it be better to have instructions of how to create the form from scratch?  Or a shared copy in Google Drive?  Will make an issue about that latter idea...

jduckles:
It would be great to have all of this information in one place. To do that we could integrate this into the workshop website and pre-populate the "workshop slug" as we have for pre and post-workshop survey links.

An example (filling in the Red sticky note text field) [Click here](https://docs.google.com/forms/d/e/1FAIpQLSd9pqNGTkEY0efaHtjOei1hsJGz8_kLrbqUt0ouug8sPExK4g/viewform?c=0&w=1&entry.1999931699=2017-03-18-someworkshop)


In the workshop template we could add a "feedback" page or link which redirects to the google form with the workshop field filled already. We just use the site.github.project-title Jekyll variable to fill in a unique value for the workshop name. 
```
https://docs.google.com/forms/d/e/1FAIpQLSd9pqNGTkEY0efaHtjOei1hsJGz8_kLrbqUt0ouug8sPExK4g/viewform?c=0&w=1&entry.1999931699={{ site.github.project_title }}
```

jduckles:
The next step would be to get the results to the instructor after the workshop.

raynamharris:
Thanks for the suggestions @ChristinaLK and @jduckles.

I agree that it would be nice to have these all in one centralized location for easy access and to facilitate  meta analyses over time. However, the instructors **must** be able to see this in real-time **during** the workshop. If we have to wait til after the workshop to get the results then we won't be able to adapt the workshop to the learners needs.

raynamharris:
Ah, I now see @ChristinaLK's new Google Drive issue, so I'll close this now as Issue https://github.com/swcarpentry/instructor-training/issues/405 will take care of this

PR #402, opened 2017-03-08
added a teaching practice discussion

raynamharris:
This particular episode does not currently have any challenges or exercise. @ErinBecker used this in a workshop and it worked really well.


ChristinaLK:
Looks good to me.

PR #404, opened 2017-03-13
Update style to v9.1.1

rgaiacs:
A few changes on style.


ChristinaLK:
I don't see anything here that will obviously break the site.  @rgaiacs, anything we should look out for specifically, or can I go ahead + merge this?

PR #412, opened 2017-03-25
add instructions for introduction

raynamharris:
I think we need to explicitly remind instructors to introduce themselves and each other.  How is this for a simple checklist @tracykteal ?

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -43,7 +43,7 @@ for a look at how using checklists can save lives (and make many other things be
 
 > ## During the Event [Trainer]
 >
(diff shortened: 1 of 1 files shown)


_extras/guide.md
```diff
@@ -43,6 +43,8 @@ for a look at how using checklists can save lives (and make many other things be
 
 > ## During the Event [Trainer]
 >
+> 1.  Introduce yourself (see detailed guid below). 
+> 1.  Ask your students to introduce themselves to each other.
```
lexnederbragt: students --> trainees (to be consistent in what we call them (I hope))

PR #416, opened 2017-03-25
adds closed captioning to instructor notes

raynamharris:
Addresses issue https://github.com/swcarpentry/instructor-training/issues/414, at least in the instructor notes

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -220,6 +220,8 @@ and have a rubric to follow.
 
 ## Live Coding Demo Videos
 
+Note on closed captioning: Sometimes audio in the room can be poor. Two suggestions for improving accessibility are to have students watch it on their own laptops and to turn on closed captioning by pressing the `cc` button at the bottom of the video.
+
 **[Part 1][live-coding-bad]: how not to do it**
 
 *   Instructor ignores a red sticky clearly visible on a learner's laptop.


maneesha:
I'm seeing this a little late.  I want to revert this PR merge.   The "live coding done poorly" does *not* have a closed captioning option.  The "live coding done right" does have a closed caption option but it is not an accurate transcription. It uses Youtube or Google's speech recognition so it does not do well with accents, jargon, or background noise.  I watched the first 60 seconds of the "done right" one with no sound, closed captioning only, and it made no sense.

I would rather not  recommend this rather than recommend something that produces basically jibberish.

lexnederbragt:
We can add our own closed captioning, apparently: https://support.google.com/youtube/answer/2734796?hl=en

maneesha:
@lexnederbragt  - yes we can but it is a lot of work to transcribe everything.  If we find someone who's willing to do it I am all for it.

ChristinaLK:
Should we email discuss and see if anyone is interested?

ErinBecker:
@ChristinaLK that sounds like a great idea. Would you like to write the email? Or maybe @maneesha? I've reverted the PR.

lexnederbragt:
As the main actor, director and producer of the videos :-) I'm OK with sending that email.

ErinBecker:
Great, thanks @lexnederbragt

PR #417, opened 2017-03-27
Add CoC to the beginning

jduckles:
Make the CoC the first thing we talk about so that people are aware of it before proceeding.


ChristinaLK:
This looks good to me.

PR #423, opened 2017-03-30
changing demo checklist

ChristinaLK:


_extras/guide.md
```diff
@@ -495,7 +495,8 @@ Checklist for instructor trainers hosting a live-coding demo session as part of
 ### After the Demo
 
 *   Sign up for your next session in <http://pad.software-carpentry.org/teaching-demos>.
-*   Add pass/fail information for the trainees in AMY.
+*   Email checkout (at) carpentries (dot) org with names + pass/fail information
+*   Clear etherpad of data from your session.
```
lexnederbragt: I leave that information in the etherpad a little while longer, in case attendees want to get the notes from there (feedback from me and the others).

ChristinaLK: My fear is that then it doesn't get cleared out.

lexnederbragt: Not an unreasonable fear. OK, I withdraw my objection :-)

PR #433, opened 2017-04-12
Add rubric suggestions

ErinBecker:
Fixes https://github.com/swcarpentry/instructor-training/issues/394

This is just to get us started. Please feel free to add or change items as this is developed.


raynamharris:
Ah. I used the brief, table version yesterday during the demo and it was lovely. With a brief glance you have some key things to look for without reading sentences. Is there anyway we can keep (some of) it and the more detailed descriptions?

ErinBecker:
@lexnederbragt @raynamharris Sure, no problem. Let's have both the rubric and the form. I'll add that back in.

PR #436, opened 2017-04-18
Removed javascript for challenges

tracykteal:
Addresses https://github.com/swcarpentry/instructor-training/issues/435

Makes it so that challenge text shows by default, rather than needing to be clicked to appear.

Removes the reference to '.challenge' in assets/js/lesson.js


tracykteal:
Sorry, added other files too. Closing and will resubmit.

PR #440, opened 2017-04-25
Point to a reference for "Lateral knowledge transfer"

arokem:
Turns out @gvwilson maybe made that term up. But Jon Udell used the term "tacit knowldege" (originally derived from Michael Polanyi: http://press.uchicago.edu/ucp/books/book/chicago/T/bo6035368.html) to refer to this exact thing. This edit still preserves the term "lateral knowledge transfer" (because it's neat!), but also points to its provenance in the link that was added.

### DIFF
--- a/_episodes/13-live.md
+++ b/_episodes/13-live.md
@@ -49,8 +49,7 @@ Its advantages are:
 *   It enables instructors to be more responsive to "what if?" questions.
     Where a slide deck is like a railway track,
     live coding allows instructors to go off road and follow their learners' interests.
-*   It facilitates lateral knowledge transfer:
-    people learn more than we realized we were teaching by watching *how* instructors do things.
+*   Lateral knowledge transfer: live coding facilitates the transfer of [tacit knowledge](http://jonudell.net/udell/2006-09-19-screencasting-of-tacit-knowledge.html) -- people learn more than we realized we were teaching by watching *how* instructors do things.
 *   It slows the instructor down:
     if she has to type in the program as she goes along,
     she can only go twice as fast as her learners,


gvwilson:
Pretty sure I didn't make it up - I'm not that smart.  But LKT is different from Polanyi's tacit knowledge: the latter is "things that are hard to transfer intentionally in written or verbal form, while the former is things that are transferred unintentionally (and often unwittingly) during demonstration of something else.  I'll keep digging...

arokem:
But the example given in that link is pretty spot on for LKT, isn't it? I mean, this is not really something that is tacit, in that a knowledgeable practitioner would be able to say "Oh yeah - the underscore: that's a fixed reference to the *last* evaluation. Good thing you noticed that!" or some-such.

ErinBecker:
A more common term for what's referred to here as "lateral knowledge transfer" is "adaptive transfer" or "far transfer" (as opposed to "analogous" or "near" transfer). Terminology like this is a pain to track down because the ed community (more than most) doesn't use standardized jargon.  

Unfortunately, the term "knowledge transfer" is also used in terms of "sharing of knowledge from knowledge producers to knowledge consumers" or "from experts to novice practitioners". Two examples of this usage are [here](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC2690353/) and [here](https://systematicreviewsjournal.biomedcentral.com/articles/10.1186/s13643-015-0094-3).

ErinBecker:
Hmm . . . actually no, I'm wrong. "Lateral knowledge transfer" here is not the same as adaptive transfer. Adaptive transfer is the use of information in situations different from the situations in which the knowledge was gained (e.g. applying debugging skills to a different type of error message). 

The PR from Ariel looks good to me. I'll go ahead and merge.

PR #445, opened 2017-05-01
Added reference to "Moonwalking with Einstein".

DataWookie:
Sitting in instructor training in Cape Town. I just read this book and thought that it was relevant to the "Expertise and Memory" session.


ChristinaLK:
Thanks Andrew!  This looks good to me.  @ErinBecker ?

ErinBecker:
Thanks @DataWookie!

raynamharris:
Thanks for this suggestion. I checked this book out from the library and I'm already halfway through reading it. So good!

DataWookie:
Great. It's a really good read. I learned some interesting tricks (?) too, some of which I have put into practice.

PR #451, opened 2017-05-07
Add reminder to have participants complete application form

maneesha:


ErinBecker:
@maneesha Thanks for adding this. Would you mind please changing the language to be directed at trainees (rather than Trainers) to keep it consistent with other language in this module and others?

Thanks!

PR #452, opened 2017-05-07
remove "two day" from DC description (it's not a difference)

maneesha:
---

### DIFF
--- a/_episodes/21-carpentries.md
+++ b/_episodes/21-carpentries.md
@@ -66,7 +66,7 @@ Accordingly, Data Carpentry workshops:
 
 *   are aimed at pure novices,
 *   domain-specific, and
-*   present a full two-day curriculum centered around a single data set.
+*   present a full curriculum centered around a single data set.
 
 Software Carpentry workshops focus on best practices for software development and use.
 Its workshops are:


maneesha:
Following venn diagram image should also be fixed accordingly.

raynamharris:
I'm not sure who created the venn diagram. This particular venn diagram point is making that DC cirriculum is centered around a single dataset and the SWC is modular. I think if we keep the existing Venn it will be fine for now, unless someone wants to completely revamp. it. 

https://github.com/swcarpentry/instructor-training/blob/gh-pages/_episodes/21-carpentries.md

ErinBecker:
I've always interpreted this the way @raynamharris mentioned - that it's a pointer towards the SWC materials being modular while the DC materials are meant to be used as a full set (for a particular workshop type).

ChristinaLK:
So is it okay as is?  I think I'm the original creator of that one.

maneesha:
I agree as @ErinBecker said above, we should make clear that the  SWC materials are  modular while the DC materials are meant to be used as a full set.

I don't want us to explicitly list the fact that it's a *two day* workshop as a *difference*  --- since SWC and DC workshops are both two days that may cause confusion.

PR #475, opened 2017-05-22
fix formatting of challenge in cognitive load lesson

fmichonneau:
an extra `>` was throwing out the parser to format the last challenge correctly


ChristinaLK:
Thanks @fmichonneau !

PR #489, opened 2017-06-29
remove language about creating an account

maneesha:
Right now all they are doing is submitting an application. Creating an account with the ability to log in to your own profile is something we'll set up after they become certified instructors.

This lesson is currently being prepared for publication and is under development at https://github.com/ErinBecker/instructor-training-reorg/. Please submit any issues or PRs to that repo (not here).


ErinBecker:
Thanks @maneesha - this is something a lot of trainees were confused by.

Issue #459, opened 2017-05-07
Practice with infrastructure: need to include DC and SWC


maneesha:
In the Carpentries module Setting up section -
We have a Practice with SWC infrastructure exercise.  This should be changed to practice with SWC or DC and include links to both program's templates.

amoskane:
Hi, I made a commit for this fix but when I try to push my branch I get "remote: Permission to swcarpentry/instructor-training.git denied to amoskane."

mkuzak:
@amoskane thank you for your eagerness to contribute to the lessons. If you would like your changes to be included, please fork the repository first, and then create a pull request with your changes. The lesson maintainers will review your contribution and merge it if they agree it fits. Please have a look at our [contribution guidelines](https://github.com/swcarpentry/instructor-training/blob/gh-pages/CONTRIBUTING.md) before considering the pull request.

PR #494, opened 2017-08-17
Practice with infrastructure: need to include DC and SWC #459

ErinBecker:


ErinBecker:
Thanks @amoskane, this looks great.

PR #497, opened 2017-09-06
bring in reorganized materials

ErinBecker:
This brings in changes from https://github.com/ErinBecker/instructor-training. 

@ChristinaLK please review and merge if everything looks good to you.

Note: I made a mistake when originally duplicating the materials into my account in that I didn't create a fork. I already had a fork of the instructor training materials that I wanted to keep separate for small-scale fixes, and couldn't have two forks of the same repo. Unfortunately, this means I couldn't do a direct PR from the reorganized materials back into this repo. I've had to create a branch here, copy over the new files and create a PR from that branch. I'm worried that this means the commit history isn't saved properly and especially worried that we may not have a correct and updated list of contributors. @ChristinaLK and others, if you know of a better way of resolving this problem, please let me know. In the future, I know that forking is important. :-/


ChristinaLK:
As long as we built the changes on a clone from the original repository, I think we can keep the history.  I'll send you an email.  I'll wait on merging this until then.

ErinBecker:
Replaced with https://github.com/swcarpentry/instructor-training/pull/498 and closed.

PR #498, opened 2017-09-06
pull in new materials

ErinBecker:
@ChristinaLK - I think this fixes the problem with the missing history. Please take a look and let me know if you see any problems. Thank you!!


ChristinaLK:
Brilliant.  I'll try to give this a look before I leave today and finalize.  :D

gvwilson:
Woo hoo!

ErinBecker:
Yep - the coffee and lunch break episodes ended up being needed to keep the schedule in sync.

I'll go ahead and merge at @ChristinaLK's approval!

PR #516, opened 2017-10-05
top 10 tips revisited

raynamharris:
I would like to point the 5 attendees of today's demo session to a "top tips/guidelines" for live coding resources. But, I see that these are hidden as a solution in the lesson materials. 

http://carpentries.github.io/instructor-training/15-live/

I would like to make these more prominent somehow. So I thought they could go at the end. I reworded them a little for clarity in this stand-alone the section. But, I think there is room for clarity or maybe we can even add a few tips....

### DIFF
--- a/_episodes/15-live.md
+++ b/_episodes/15-live.md
@@ -152,5 +152,17 @@ Remind learners frequently about using their sticky notes, or they (and you) wil
 > This exercise should take about 20 minutes.  
 {: .challenge}
 
+## Breif List of Tips for Live Coding in a Workshop
+1. Stand up and move around the room if possible. 
+1. Go slowly while you say commands as you type them and point to outputs on the screen. 
+1. Mirror your learner’s environment by turning off fancy enviornments and by not using keyboard shortcuts.
+1. Use a big font  and maximize the window with good contrast so that all learners can see
+1. Use illustrations to help learners understand and organize the material.
+1. Turn off notifications on your laptop and phone.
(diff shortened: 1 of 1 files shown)


_episodes/15-live.md
```diff
@@ -152,5 +152,17 @@ Remind learners frequently about using their sticky notes, or they (and you) wil
 > This exercise should take about 20 minutes.  
 {: .challenge}
 
+## Breif List of Tips for Live Coding in a Workshop
```
ChristinaLK: Breif -> Brief

lexnederbragt:
The Top 10 tips was a separate list in the main document before it was shortened and moved as a solution to an exercise during the reorganisation. This PR puts part of it back at the end of the main text, but then duplicates much of the solution of the exercise.

As the author if the original top 10 I may be biased, but I'd like to see the full top 10 text back in the main document. One would still ask the trainees to come up with these pojnts, and during a workshop perhaps only briefly check whether all was treated. But at least this would solve Rayna's issue with wanting something to point trainees (and others) to.

Alternatively, a lonk could be added [this blog post](https://software-carpentry.org/blog/2016/04/tips-tricks-live-coding.html) with the original, full text.

ChristinaLK:
@lexnederbragt can you make this into an issue?

weaverbel:
I just want to say that I really missed the top ten tips when I last taught in Perth @lexnederbragt   and wondered where they had gone. Would definitely be +1 for putting them back in.

PR #532, opened 2017-10-20
Fix a broken link to another part of the lesson.

arokem:
It seems to me there should be a good way to do this using templating,
but the current way this is done results in a broken link.


ErinBecker:
I think this will work.

PR #536, opened 2017-10-24
05-memory.md: python formatting, docstrings

naught101:
Use `.python` style for better formatting, use descriptive docstrings.


ChristinaLK:
good idea - thanks @naught101

naught101:
Hrm.. Looks like the .python syntax highlighting isn't enabled anyway? Which maybe means that this PR actually makes things harder to read, because the comment is more complex. With syntax highlighting, it would be clearer. Would it be possible to enable it? I haven't used Jekyll before..

PR #538, opened 2017-10-24
WorldCat link rather than link to specific seller

stragu:
Currently, the list of books by Patricia Benner is an Amazon link.
I think it would be more reasonable to link to a general catalogue of books (WorldCat seems exhaustive enough) and let the reader choose their preferred seller, rather than directly providing them with a link to a specific seller we might seem to give preference to.


ErinBecker:
@stragu - Thanks for this PR. I agree that it's a bit odd for us to point directly towards a particular vendor. As @ChristinaLK says, there are certainly other Amazon links in the materials. If you'd be willing to go through and change those links as well, it would be a big help!

stragu:
Thanks @ChristinaLK and @ErinBecker . I'll definitely keep an eye out for other links in the material I will be reviewing for instructor training, but won't commit to looking for them all just yet.
Cheers

ChristinaLK:
For now I'll open an issue as a placeholder reminder.

PR #541, opened 2017-10-24
minor correction key point

aschuerch:
I believe it must say faded example not formative assessment


ErinBecker:
@aschuerch - Thanks for the close reading. It is meant to be "formative assessment" however. The text it's referring to is in the first part of that episode:

"If we present our learners with large amounts of information, without giving them the opportunity to practice using it (and thereby transfer it into long-term memory), they won’t retain the material as well as if we present small amounts of information interspersed with practice opportunities. This is yet another reason why frequent formative assessment is important."

Issue #501, opened 2017-09-21
Changing order of PR things to do in checkout


tracykteal:
This came up on the Community Call today as a suggestion from Ashwin Srinath. Commenting on a pull request is as important as submitting your own. This is currently listed as on option in the checkout materials. However we could switch the order of what you can do for checkout to 'comment on a PR, create an issue or create an issue'. 

@ChristinaLK

PR #542, opened 2017-10-24
change order of suggestions for lesson contribution

ErinBecker:
Fixes https://github.com/carpentries/instructor-training/issues/501


ErinBecker:
@ChristinaLK @tracykteal @maneesha

tracykteal:
LGTM. I like that order.

PR #545, opened 2017-10-26
remove print statement

fmichonneau:
this is probably a left over from the translation of the lesson from python, and not needed. In R, calling a variable at the terminal defaults on the print method.


ErinBecker:
Thanks @fmichonneau - I copied this exercise over directly from one of the SWC R lessons. For reasons that I'm not perfectly clear on, they've decided to keep the `print` statements in the R code, after acknowledging that they're not necessary. I don't see any problem with removing it here, but instructors will see it in the SWC R lesson in any case.

PR #559, opened 2017-11-05
Clarify that checkout steps must be completed in order

colinmorris:
The previous wording was ambiguous. "you must complete these steps in order to complete your training" could be read as "in order to complete your training, you must complete these steps".

(An alternative solution would be to keep the previous wording but parenthesize "in order")

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -7,9 +7,9 @@ demopad: http://pad.software-carpentry.org/teaching-demos
 discussionpad: http://pad.software-carpentry.org/instructor-discussion
 ---
 
-After you have finished the two-day training class,
-you must go through these steps in order to complete your training and be
-fully certified as a Data and Software Carpentry instructor:
+After you have finished the two-day training class, to complete your training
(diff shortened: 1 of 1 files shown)


colinmorris:
Actually, now I'm doubly confused. The rest of the document supports the interpretation that the steps have to be done in order (part 2 begins "After you have submitted a contribution to one of our lessons, the next step is to take part in an hour-long discussion session led by an experienced instructor. ").

But I just noticed that the Data Carpentry [checkout instructions](http://www.datacarpentry.org/checkout/) say:

> Tasks are listed in the order most of our instructor trainees complete the checkout process, but you can complete them in any order.

So I'm actually not sure which one reflects reality.

I'll ask the checkout helpline for clarification and close this P.R. for now.

ChristinaLK:
@colinmorris, the order doesn't matter (although it usually makes sense to do the demo last) so we should change the wording in the checkout page of the training material.

lexnederbragt:
Addressed in #560 and #561

PR #560, opened 2017-11-06
Clarify that checkout steps may be completed in any order

colinmorris:


ErinBecker:
Thanks @colinmorris! This really helps to clarify.

ChristinaLK:
Yes, thank you!

Issue #567, opened 2017-11-16
Update code of conduct contact email addresses


nspies:
The instructor training ["teaching practices"](instructor-training/_episodes/22-practices.md) section lists "admin@software-carpentry.org or admin@datacarpentry.org" as contact addresses to report code of conduct violations.

I believe the instructor materials should be updated with the new @carpentries.org email addresses in the [code of conduct itself](https://software-carpentry.org/conduct/), which says that code of conduct violations should be directed to "the joint Data/Software Carpentry (“the Carpentries”) Policy subcommittee by emailing policy@carpentries.org or C. MacDonnell at confidential@carpentries.org".

ErinBecker:
Hi @nspies - thanks for your attention to detail. There are a couple of different reporting types involved in CoC violations. For online incidents people should contact the Policy subcommittee, or if for some reason they'd prefer not to contact them, the confidential@carpentries.org email is an alternate. For cases where the incident occurred at an in-person event, the immediate contact should be the workshop host and/or coordinator. Because the instructor training materials are targeted towards instructors who will be leading in-person events, the workshop coordinator (and/or host) is the appropriate person to contact. The relevant part of the CoC is: 

>> If you believe someone is violating the Code of Conduct we ask that you report it to the joint Data/Software Carpentry (“the Carpentries”) Policy subcommittee by emailing policy@carpentries.org or C. MacDonnell at confidential@carpentries.org, and if the violation occurs during a workshop or other in-person event, by contacting the host and/or coordinator.

I agree that this wording may be somewhat confusing, but it's not a case of forgetting to update the email addresses! If you have a suggestion for making the text of the instructor training materials more clear, please let me know. I'd be happy to implement.

nspies:
Hi @ErinBecker - I think there may have been some confusion about which section in the instructor training I was referring to so I've made a pull request that hopefully clarifies (sorry, this was me trying to do a quick issue report during the relevant part of our training!).

The section in question directs the workshop coordinators to contact those email addresses as follow-up to the workshop and so I do think the instructor materials should include the new email addresses.

ChristinaLK:
This has been resolved by #572, correct?

nspies:
Yes!

PR #572, opened 2017-11-20
Update emails for reporting code-of-conduct violations

nspies:
This updates the relevant section of the instructor training with the new `@carpentries.org` email addresses for follow-up to code-of-conduct violations occurring during a workshop. (https://github.com/carpentries/instructor-training/issues/567)

### DIFF
--- a/_episodes/22-practices.md
+++ b/_episodes/22-practices.md
@@ -87,12 +87,15 @@ about how a Carpentry workshop operates.
 If you are an instructor, and believe that someone in a workshop has
 violated the Code of Conduct, you may warn them, ask them to
 apologize, and/or expel them, depending on the severity of the
-violation and whether or not you believe it was intentional.  
-You can also bring in the workshop host to mediate an incident
-if you are uncomfortable or unable to do so yourself.
+violation and whether or not you believe it was intentional.
 
-No matter what you choose to do, please contact the joint Data/Software Carpentry Policy subcommittee by emailing [policy@carpentries.org](mailto:policy@carpentries.org) or C. MacDonnell at [confidential@carpentries.org](mailto:confidential@carpentries.org) as
-soon as you can and let us know what happened so that we can follow-up
+The local workshop host/admin is expected to help enforce the Code of Conduct and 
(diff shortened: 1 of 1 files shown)


ErinBecker:
Thanks @nspies for putting in this PR. I think this actually raises an interesting question about our reporting procedure. Thus far, all reported instances of potential CoC violations have been for conduct that occurred online and have been reported through the policy email you reference here. However, for workshops, we want Instructors to (at least) notify the workshop administrator of the issue, even if they don't feel it warrants a formal report to the Policy subcommittee. Even if they are submitting an official report, we want Instructors to also let the workshop host and/or administrator know what happened, in case there is immediate action that is required. Could you please update your PR to include the original wording and add a sentence about reporting the incident to the policy committee?

nspies:
I'm not sure I'm following everything. It sounds like you're distinguishing a local host/administrator from the instructor who's running/helping with the current lesson? And you'd like the instructor training to first direct the instructors to report issues to the host/admin, during the workshop, and afterwards follow-up with a report to the carpentries emails?

ErinBecker:
@nspies - Yes, that's correct. For workshops that are organized by the Carpentries, there is a workshop administrator (usually Carpentry staff) and a workshop host (the local contact at the organization who handles things like catering and room reservations). We want to make sure that the instructor teaching the workshop notifies these people so that any immediate action that's appropriate can be taking by the host (who is on site) and so that any longer-term follow up that is necessary may be handled by the admin. In an extreme example, if the CoC violation created an unsafe environment for the instructor and the host was not responsive to the problem, we want the administrator to be aware so that we can take that into account when deciding whether to send instructors to that site again. 

So we need to keep the instructions to the instructor to notify the admin and/or host, but we can certainly also add your wording about notifying the policy committee after the incident as well.

Hope this clarifies.

nspies:
Okay, thanks, that's clarified things for me and I've adjusted the language so it's hopefully clearer to trainers/instructors as well.

(I'll note that the current language doesn't really suggest that instructors should settle an issue themselves without reporting it to hq.)

ErinBecker:
Thanks @nspies!

PR #574, opened 2017-11-27
Updated issues links

nkindlon:
The [issues] link, defined at the bottom of the page, needed to be updated from:
https://github.com/swcarpentry/instructor-training/issues/
to:
https://github.com/carpentries/instructor-training/issues/


ErinBecker:
@nkindlon - Thanks for putting this through so quickly. My appologies but I misunderstood your original issue. It is only the one link that needs to be changed: https://github.com/swcarpentry/instructor-training/ --> https://github.com/carpentries/instructor-training/. The styles repo and workshop template repo are still in the old location. I'll update your PR with these changes and merge.

PR #575, opened 2017-12-02
Add diagram for skill acquisition

yochannah:
I've added an illustration to demonstrate the different levels of experience from Novice through Competent Practitioner and Expert. You can preview it here: 

https://yochannah.github.io/instructor-training/02-practice-learning/


katrinleinweber:
Hello! I don't have merge rights here, but because some "remove trailing whitespace" setting in your editor seems to have enlarged the diff here a lot, I would recommend to revert `d782627` and re-do it without that setting. That would make it easier in the future to as per the https://swcarpentry.github.io/git-novice/04-changes/ > "good commit messages" reference.

yochannah:
I literally had to run a diff with the -w command to suppress all the whitespace and didn't think about the fact it might bother others. 🤦‍♀️

yochannah:
new, better PR: https://github.com/carpentries/instructor-training/pull/578

katrinleinweber:
Cool thanks! Were you using GitHub's web-interface? Because it can now squash commits before merging, so it's not an issue to push reverts & more commits into the same branch and thus PR ;-)

yochannah:
I wasn't, actually! I tend to fall on the side of preferring not to squash/rebase (I've seen some really messed up rebases) so sometimes forget it exists and could be useful in scenarios like this.

PR #580, opened 2017-12-06
Removed unnecessary SWC/DC distinction

lexnederbragt:


ErinBecker:
@lexnederbragt - We actually do still want to have this information. Our system is still set up to record people as licensed DC or SWC instructors and the certificate that we send them makes this distinction. Many of the Trainees I've talked with care about which of the certificates they get. Thanks for putting this in, but given that we still need to record this information, I'm going to go ahead and close this PR.

Issue #511, opened 2017-09-28
Zoom breakout rooms for distributed training - be sure to exclude co-instructor/helpers


maneesha:
Zoom instructions should be updated to note that instructors and helpers should be excluded from breakout rooms.

PR #585, opened 2017-12-15
update zoom instructions

ErinBecker:
Re: https://github.com/carpentries/instructor-training/issues/511


ErinBecker:
Thanks @karenword, I've made that modification and merged.

Issue #539, opened 2017-10-24
The Unix shell lesson as an example for "going slowly" is outdated


stragu:
The chapter "[How learning works: the importance of practice](https://carpentries.github.io/instructor-training/02-practice-learning/)" uses the Unix shell lesson as an example of "going slowly". It currently states:

> As an example of this idea’s practical implications, Software Carpentry’s lesson on the Unix shell only introduces 16 commands, and we usually spend two and a half hours working through it. Eleven minutes per command may seem glacially slow, but the lesson’s real purpose is to teach learners about paths, history, wildcards, pipes and filters, command-line arguments, redirection, and all the other big ideas on which the shell depends, and without which people cannot understand how to use commands.

However, the [Unix shell lesson](http://swcarpentry.github.io/shell-novice/) currently teaches about 21 commands, not 16:

```
cd
whoami
pwd
ls
mkdir
nano
rm
mv
cp
wc
cat
sort
head
echo
uniq
cut
grep
tail
bash
history
find
```
(this list does not include tricks like `PS1=xxx`)

Furthermore, the recommended time spent on it is 1h35, not 2h30, which makes it 4.5 minutes per command (as opposed to 11 minutes).

Is this still a good example of "going slow"? If it is, the numbers should simply be updated.
If not, what should we do?

**Edit**: oops, counted `cd` twice -> 21 commands instead of 22.

ChristinaLK:
I think the times are off for the shell lesson, because it should still take about 2.5 hours.  Good catch on the number of commands though -- we should update the numbers.

ErinBecker:
@stragu - Thanks for catching this! I did update the the number of commands and the calculation for this a couple of months ago when we published the new versions of the lessons, but it must have changed again! 

I'm not sure the best thing to do here. The shell lesson should definitely take more than 1h35 (and has taken about 2.5 hours each time I've taught it), but I don't want to submit a bunch of timing updates to the shell lesson and drag out the conversation of which episodes take longer. 

I think our options are:

1) Keep the reference to 2.5 hours in our lesson and ignore the fact that that doesn't line up with recommended timing in the shell lesson. Update the total number of commands to 22 and say that's about 6.5 minutes per command.

2) Revise the time estimate in our lesson to 1h35 minutes. Update the total number of commands to 22 and say that's about 4 minutes per command. (I don't favor this option as it might lead instructors to speed through the shell lesson.)

3) Change to a different example. 

[...]

PR #588, opened 2017-12-15
update timings for unix commands

ErinBecker:
Re: https://github.com/carpentries/instructor-training/issues/539


ChristinaLK:
@karenword , I agree w/ you!  6.5 min per command seems about right.  :)

ErinBecker:
Agreed! 6.5 minutes per command is not terribly slow. I'd be happy to remove those words if you want to put in a rewording.

PR #601, opened 2018-01-11
Fixes another broken link in welcome.txt

arokem:


remram44:
Some of the links down here appear broken too: http://carpentries.github.io/instructor-training/guide/#not-good-starting-points-for-demos

PR #609, opened 2018-01-23
Use correct URL for SWC books

njtierney:
This pull request links the text

> If you are interested in doing more reading, we have some recommendations

to this link: https://carpentries.github.io/instructor-training/reference

Previously, this link was https://carpentries.github.io/instructor-training/reference#Books which did not work


rgaiacs:
If you use https://carpentries.github.io/instructor-training/reference#books it will work.

PR #610, opened 2018-01-26
removed instructor instructor names

raynamharris:
added new instructor trainers. See also https://github.com/carpentries/trainers/pull/7


ChristinaLK:
So this trainer list is being removed from this repository because that information is being stored in the "trainers" repo?

raynamharris:
Yes @ChristinaLK that is the idea

ChristinaLK:
I'm okay with deleting here, just wanted to make sure we aren't losing the information.  :)

Issue #611, opened 2018-01-30
link to worksheet in How We Operate lesson is broken


amyehodge:
This currently points to https://carpentries.github.io/instructor-training/files/handouts/Carpentries_roles_worksheet.docx, which does not exist.

This needs to point to https://carpentries.github.io/instructor-training/files/handouts/Carpentries_roles_worksheet_v2.pdf

katrinleinweber:
Hi :-)

This can be closed then, right?

Unless there is a reason in keeping this open after the corresponding PR is merged, you can use [keywords like `fixes #…` in the commit message ](https://help.github.com/articles/closing-issues-using-keywords/) to make GitHub close issues automatically upon merging the PR ;-)

amyehodge:
fixed

amyehodge:
Thanks for the tip! I’ve closed it.

~ Amy

Amy E. Hodge, PhD
Science Data Librarian
amyhodge@stanford.edu<mailto:amyhodge@stanford.edu>
650.556.5194
[cid:image001.png@01D39A64.2DECFF30] orcid.org/0000-0002-5902-3077<https://orcid.org/0000-0002-5902-3077>

Data Management Services
Branner Earth Sciences Library, 212 Mitchell
397 Panama Mall; MC 2211
Stanford University
Stanford, CA 94305

PR #612, opened 2018-01-30
fixing broken worksheet link

amyehodge:
This takes care of issue #611 

If someone could merge this before our training on Feb. 5-6, that would be greatly appreciated.


ErinBecker:
Thanks @amyehodge!

PR #620, opened 2018-02-05
Add links to labelled GitHub issues in checkout

colinmorris:
Direct would-be contributors to queries for issues with tags like `newcomer-friendly`, `discussion`, etc. (across all carpentries repositories).

Something I noticed while putting these queries together: it seems like the swcarpentry and datacarpentry orgs use slightly different labels. For example, `newcomer-friendly` and `help-wanted` in swc, but `beginner-friendly` and `help wanted` in dc ([example](https://github.com/datacarpentry/python-ecology-lesson/issues/232)). Is there a good place to report this?

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -56,9 +56,11 @@ with our collaborative curriculum development process.
 
 Trainees must make a contribution to a lesson's content, exercises, or instructor's guide by doing one of the following:
 
-1. Providing substantive feedback on an existing issue or pull request (preferred). 
-2. Submitting a change to fix an existing issue.
-3. Proof-reading a lesson and adding a new issue describing something to be improved.
+1. Providing substantive feedback on an existing issue or pull request. A good starting point may be issues and pull requests with the label [`discussion`](https://github.com/search?utf8=%E2%9C%93&q=org%3Aswcarpentry+is%3Aopen+org%3Adatacarpentry+label%3Adiscussion) or [`question`](https://github.com/search?utf8=%E2%9C%93&q=org%3Aswcarpentry+is%3Aopen+org%3Adatacarpentry+label%3Aquestion).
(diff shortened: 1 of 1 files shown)


colinmorris:
Oh, I should also mention that this goes from marking 1 (feedback on existing issue/PR) as the preferred form of contribution to 1 and 2 (a PR addressing an existing issue). My thinking was that the main thing we wanted to discourage was unsolicited PRs, but that *solicited* PRs (particularly if they're in response to an issue a maintainer has labeled as `help-wanted`) are encouraged. 

But I'm happy to package that part into a separate issue/PR if you think it bears more discussion.

katrinleinweber:
+1 for using SWC labels, because "newcomer" also includes subject matter experts, who may be new to the Carpentries, while "beginner" may be understood as only new to the subject itself.

ChristinaLK:
@colinmorris there's been discussion of dropping that particular requirement, so it may not matter so much for checkout.  
However I am +1 for standardizing tags.  I'll flag @ErinBecker as she's currently spending a lot of time w/ the maintainer community and can say where this should go next.

fmichonneau:
As an update, we are working on a proposal to harmonize the github issue labels across repositories. Please send your comments and feedback here: https://docs.google.com/document/d/1b3nIZ6N4IHY24JmLNQ5rkwUACEVS9Hls3auzZD7zHqk/edit# Thanks!

katrinleinweber:
In addition to my [earlier comment](https://github.com/carpentries/instructor-training/pull/620#issuecomment-363343213) please do link to the `help wanted` and `good first issue` labels. This PR can be updated or re-submitted.

I agree with François that nothing with `beginner` or `newcomer` should be used after all.

Even if people don't _have to_ do that, funneling people to issues exactly labelled for them has a good chance of supporting the maintainers. The [current text](https://github.com/carpentries/instructor-training/blob/a54b8f27d151b8b775fd02f5f8d26ac126049bc5/_episodes/23-checkout.md#instructor-checkout) give no pointers at all :-(

However, it might be more useful to update that text after the repos have all moved to github.com/carpentries, so the 6 possible link targets are reduced to 2.

PR #624, opened 2018-02-14
punctuation, another tip and save chat

weaverbel:
Hi Karen feel free to reject my suggested changes


karenword:
Thanks @weaverbel!

PR #625, opened 2018-02-14
Added XKCD's imposter syndrome cartoon

lexnederbragt:
If the image needs to be added as a png file instead of embedding it as a link, let me know and I'll fix that.


karenword:
I'd love to have the image, but am unsure about licensing in that regard. Linking seems safe and worthwhile! Thanks @lexnederbragt

PR #633, opened 2018-02-28
remove link to DC workshop-template

fmichonneau:
Now that the Data Carpentry workshop template repository is deprecated, I removed mention of it in this lesson. I also removed "SWC" to avoid confusion. The workshop template is the same for all types of workshops (SWC, DC or LC). There is a variable to set in the template to get the type of workshop needed.


karenword:
Thanks @fmichonneau !

PR #637, opened 2018-03-05
Removed duplicate bullet

kariljordan:
Improvement-based praise was listed twice


karenword:
These bullets are meant to provide solutions to each of the 4 bulleted quotes above so  "improvement-based praise" is the correct classification for both the second and 4th quotes. Closing this PR, but feel free to submit an alternative that makes this matching more clear!

PR #648, opened 2018-03-19
Fix license link.

Petraea:
Hi there,
The licensing link was broken on this page. This is a simple fix to correct it.


karenword:
Thanks @Petraea !

PR #650, opened 2018-03-21
Changed "Your Academic Past" for readailbity.

loz-hurst:
Changed single paragraph with long list of questions as single block of text to bulleted list of questions.  I find this much more readable and easier to access.


ChristinaLK:
Good change @loz-hurst.  Thank you!

PR #660, opened 2018-03-28
add slack badges

fmichonneau:
Add slack badges to join organization and lesson specific channel on top of README


ErinBecker:
Thanks @fmichonneau !

PR #661, opened 2018-04-02
Fixed typo

kariljordan:


ChristinaLK:
I don't think this is a typo -- it's a parallel construction (avoiding A+B in favor of C+D).

PR #663, opened 2018-04-03
Fix hyperlink

katrinleinweber:
See diff ;-)


karenword:
Thanks @katrinleinweber !

katrinleinweber:
You're welcome, and thanks for merging :-)

Issue #616, opened 2018-01-31
The Importance of Mindset Quiz - New Suggestion


goldmj:
Have an issue with the recommended "official mindset self-assessment": https://mindsetonline.com/testyourmindset/step1.php.
This quiz uses "subjective" language and the format turns quiz takers off by repeating questions, and having lot of pages to click through. Also, the "Buy My Book" at the end and around the website is off-putting. Think product placement should be avoided in this open source curriculum. 

Suggest switching to: http://blog.mindsetworks.com/what-s-my-mindset (or another option!)
This quiz breaks down the questions into more comprehensible language, is one page, and maybe shorter for people to stick with and fill out.

amyehodge:
I'm going to use this one http://www.londonacademyofit.co.uk/learning-blog/learning/interactive-quiz-fixed-vs-growth-mindset/ when I teach this next week. It's similar to the 8 question version that Julie points to above, but has a nice slider that moves as you answer the questions. I agree that the 16-question version that we link to feels VERY repetitive, though I suspect it was chosen because it's Carol Dweck's site for this book, which we are using extensively in this lesson.

kariljordan:
I think the mindset quiz @goldmj suggested is straight to the point and the page doesn't have a lot of "other stuff" going on that would distract the learners.

goldmj:
Another great example @amyehodge! I like the introductory image on the site, and 'keep it up' mentality at the end of the quiz!

But, wonder if the slider would introduce any bias/distract the user as they take the quiz? And also worry about other videos/distractors @kariljordan brings up.

gvwilson:
[This article](https://educhatter.wordpress.com/2017/03/26/growth-mindset-is-the-theory-flawed-or-has-gm-been-debased-in-the-classroom/) is a good summary of challenges to mindset theory and ways in which it is being mis-used - if were going to cover mindset in workshops, should we be pointing at these reservations and qualifications as well?

ChristinaLK:
@gvwilson can you open a separate issue w/ that link?

ChristinaLK:
Does taking this quiz add to learner's understanding of the topic?  If not, I would move that we simply remove it.

amyehodge:
@ChristinaLK I would argue that this quiz does not add to their understanding. It feels more like a way to include an activity.

ErinBecker:
I agree with removing it.

PR #666, opened 2018-04-12
Replace mindset quiz with think-pair-share

karenword:
Per discussion in https://github.com/carpentries/instructor-training/issues/616


karenword:
Changed in etherpad also

Issue #544, opened 2017-10-25
replace amazon links


ChristinaLK:
Per this PR (#538) it makes sense to have more general links to recommended books instead of a specific seller.

ErinBecker:
Related PR: https://github.com/carpentries/instructor-training/pull/583

ErinBecker:
This shows all places in the lesson files where we link to Amazon: https://github.com/carpentries/instructor-training/search?utf8=%E2%9C%93&q=amazon&type=

karenword:
Looks like these are all done! Thanks @murraycadzow !

PR #667, opened 2018-04-13
Amazon to worldcat

murraycadzow:
in reference to issue https://github.com/carpentries/instructor-training/issues/544 I have gone through and updated the amazon links to now point to WorldCat


ChristinaLK:
YAY!  Thank you so much @murraycadzow.

Issue #657, opened 2018-03-27
Including demo rubric in Live Coding lesson


karenword:
The rubric that is provided to Trainers to evaluate checkout demonstrations isn't provided to learners to evaluate each other. Is there any reason for this, or should we add it to the Live Coding lesson?
https://github.com/carpentries/instructor-training/blob/gh-pages/files/teaching-demo-rubric.md

lexnederbragt:
There is no reason not to add it. +1 to adding from me.

PR #692, opened 2018-05-14
Clean up demo rubric for addition to curriculum

karenword:
Step 1 to address https://github.com/carpentries/instructor-training/issues/657

This PR is intended to prepare the demo rubric for use within the instructor training curriculum, as well as making it more useful to Trainers, particularly those who are new to running teaching demonstration sessions. 

This change:
- adds information about low failure rate to counter fears of trainers and trainees, most of whom have been traumatized by high-stakes testing. 
- simplifies language in the rubric to make it easier to read 
- adds positive actions to two negatives that previously lacked clear opposites
- modifies the note-taking form to directly and sequentially follow the rubric

### DIFF
--- a/files/teaching-demo-rubric.md
+++ b/files/teaching-demo-rubric.md
@@ -24,7 +24,7 @@ Uses notes on paper/tablet|Displays lesson materials/notes on their screen|
 -|Uses significant amounts of dismissive language (e.g. "just", "easy")|
 Speaks and types at an appropriate (slow enough) pace|Speaks or types very quickly|
 Speaks clearly|-
-Appears focused and attentive to learners|Appears significantly distracted|
+Appears focused and attentive to learners|Appears significantly distracted by content or process of consulting notes|
 
 ---
 
--- a/files/teaching-demo-rubric.md
+++ b/files/teaching-demo-rubric.md
@@ -45,7 +45,7 @@ Date:
 |<br>Embraces & uses mistakes|
 |<br>Use of notes on paper/tablet|
 |<br>Distraction-free screen|
-|<br>Dismissive language|
+|<br>Wording|
 |<br>Pacing|
 |<br>Clarity|
 |<br>Demeanor & focus|
--- a/files/teaching-demo-rubric.md
+++ b/files/teaching-demo-rubric.md
@@ -3,28 +3,28 @@ This rubric is provided as a guide for Trainers evaluating potential new instruc
 Trainer. As such, deviation from this rubric is encouraged as needed to accurately assess the trainee's preparation and instructional
 skills.
 
-In general, trainees who have three or more marks in the "Negative" columns below should be asked to redo their demonstration. Even one 
-mark in a "Negative" column can be justification for asking a trainee to redo their demonstration if the problem is significant. As always,
+In general, these demonstrations have a very low (<5%) failure rate. When a trainee is asked to redo a demo, it is simply a matter of correction -- an error to embrace and to learn from -- not rejection. Trainees who have three or more marks in the "Negative" columns below should probably be asked to redo their demonstration. Even one 
+mark in a "Negative" column can be considered justification for asking a trainee to redo their demonstration if the problem is significant. As always,
 Trainers should use their own judgement when applying this rubric in individual cases.  
 
 |Positive Content|Negative Content|
 |------|---------------------|
-Trainee uses the official curricular materials as written or with minor deviations|Trainee deviates significantly from the official curricular materials|
-Trainee places content in context and explains why it is relevant/useful to the learner|Trainee jumps into the content without explaining context|
-Trainee obviously knows the material they are teaching|Trainee makes factual errors in teaching the content and doesn't correct themselves|
+Uses official curriculum with only minor deviations|Deviates significantly or doesn't use official curriculum|
+Places content in context and explains relevance/utility to learners|Jumps into the content without context|
+Teaches content correctly|Makes factual errors in content and doesn't correct|
 
 |Positive Delivery|Negative Delivery|
 |------|---------------------|
-Trainee mirrors the learner's environment (e.g. default terminal setup, simplified shell prompt)|Trainee uses shortcuts or tools not immediately available to or familiar to the learner|
-Trainee uses appropriately sized fonts and windows|Font or windows are too small|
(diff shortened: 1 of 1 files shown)


files/teaching-demo-rubric.md
```diff
@@ -36,16 +36,16 @@ Date:
...
+|<br>Context|
+|<br>Knowledge|
+|<br>Mirror learner's environment|
+|<br>Font & window size|
+|<br>Explanation of typing|
+|<br>Embraces & uses mistakes|
 |<br>Use of notes on paper/tablet|
-|<br>Avoid Being Disturbed|
+|<br>Distraction-free screen|
+|<br>Dismissive language|
```
lexnederbragt: Perhaps remove the word 'dismissive'.  It is the only negative word and really stands out. The other elements are all neutral or positively worded

files/teaching-demo-rubric.md
```diff
@@ -3,28 +3,28 @@ This rubric is provided as a guide for Trainers evaluating potential new instruc
 Trainer. As such, deviation from this rubric is encouraged as needed to accurately assess the trainee's preparation and instructional
 skills.
 
-In general, trainees who have three or more marks in the "Negative" columns below should be asked to redo their demonstration. Even one 
-mark in a "Negative" column can be justification for asking a trainee to redo their demonstration if the problem is significant. As always,
+In general, these demonstrations have a very low (<5%) failure rate. When a trainee is asked to redo a demo, it is simply a matter of correction -- an error to embrace and to learn from -- not rejection. Trainees who have three or more marks in the "Negative" columns below should probably be asked to redo their demonstration. Even one
```
lexnederbragt: These points are really good to add!

lexnederbragt: How about "these demonstrations have a very high (>95%) pass rate" -> phrasing it positively.

amyehodge: Framing is very important here. This framing will cause far less anxiety.

lexnederbragt:
I did a teaching demo yesterday and used the new form to take notes. It works very well! One questions is what the item `knowledge` is meant to be used for...

ChristinaLK:
This looks great to me -- thanks Karen!

files/teaching-demo-rubric.md
gvwilson: - would "have a very low failure rate (less than 5%)" be easier for non-native English speakers to parse?
- rather than "simply a matter of correction", prefer "usually a matter of correction"
- rather than "probably be asked to redo their demonstration", "probably be asked to try again" (to avoid repeating "redo")

files/teaching-demo-rubric.md
```diff
@@ -3,28 +3,28 @@ This rubric is provided as a guide for Trainers evaluating potential new instruc
...
+mark in a "Negative" column can be considered justification for asking a trainee to redo their demonstration if the problem is significant. As always,
 Trainers should use their own judgement when applying this rubric in individual cases.  
 
 |Positive Content|Negative Content|
 |------|---------------------|
-Trainee uses the official curricular materials as written or with minor deviations|Trainee deviates significantly from the official curricular materials|
-Trainee places content in context and explains why it is relevant/useful to the learner|Trainee jumps into the content without explaining context|
-Trainee obviously knows the material they are teaching|Trainee makes factual errors in teaching the content and doesn't correct themselves|
+Uses official curriculum with only minor deviations|Deviates significantly or doesn't use official curriculum|
```
gvwilson: - The word "official" seems pretty formal - "standard", maybe?

files/teaching-demo-rubric.md
```diff
@@ -3,28 +3,28 @@ This rubric is provided as a guide for Trainers evaluating potential new instruc
...
-Trainee speaks and types at an appropriate pace|Trainee speaks or types very quickly|
-Trainee speaks clearly|-
--|Trainee appears significantly distracted|
+Mirrors the learner's environment (e.g. default terminal setup, simplified prompt)|Uses shortcuts or tools that are unfamiliar/unavailable to learners|
+Uses appropriately sized fonts and windows|Font or windows are too small|
+Explains all typing|Types commands or troubleshoots errors without explaining|
+Uses mistakes/typos as opportunities for learning|Ignores or dismisses mistakes/typos that they make|
+Uses notes on paper/tablet|Displays lesson materials/notes on their screen|
+-|Notifications show up on their screen|
+-|Uses significant amounts of dismissive language (e.g. "just", "easy")|
```
gvwilson: - Would add something about consulting the notes too often - I've had demos in which the instructor was checking every few seconds.
- Additionally, should there be an item for "seems to know the material well" vs. "seems unfamiliar with the material"? (the latter may have inspired the frequent note checking mentioned above, or that may have been a nervous tic)

files/teaching-demo-rubric.md
kariljordan: +1

files/teaching-demo-rubric.md
karenword: @lexnederbragt would "dismissive-free language" work here, mirroring the criterion above? This piece also doesn't have a clear positive in the elaborated rubric, and I'm struggling to come up with one. We don't need to substitute dismissive language with mandatory positives, we need clear speech that lacks these kinds of "helpful" (but not actually helpful) decorations. Substitutions with more effectively encouraging phrases are at the instructor's discretion and don't really need to be scored.

lexnederbragt: What about 'Wording' - which can be scored as positive and negative during the teaching demo?

ChristinaLK:
@karenword is going to test changes at next training before merging.

karenword:
I've now field-tested this on several occasions as a replacement for the 2x2 grid during the 2nd round of Practice, with entirely positive feedback. (I have also stopped receiving complaints that they don't get direct feedback from Trainers during practice sessions, so I think this meets that perceived need for more guidance.) I've made a couple more tweaks based on comments here & will start working on a PR that incorporates this rubric into that second practice exercise.

kariljordan:
Would you like to merge now, or wait?

karenword:
Thank you for asking @kariljordan ! For this one in particular, since it's not actually in the curriculum, we can merge now. :) All other non-typo PRs should wait.

PR #699, opened 2018-05-28
Updated to include Library Carpentry

libcce:
Updated to include Library Carpentry per recent Carpentries/LC merger agreements and @tracykteal, John Chodacki.

### DIFF
--- a/_episodes/20-carpentries.md
+++ b/_episodes/20-carpentries.md
@@ -20,20 +20,23 @@ about how you can get involved.
 
 ## History
 
-[Software Carpentry]({{ site.swc_site }} ) was founded in 1998 and 
-[Data Carpentry]({{ site.dc_site }}) was founded in 2013. You can learn more about the history
-and goals of each organization by reading 
-"[Software Carpentry: Lessons Learned](http://f1000research.com/articles/3-62/v2)" and
-"[Data Carpentry: Workshops to Increase Data Literacy for Researchers](http://ijdc.net/index.php/ijdc/article/view/10.1.135)".
+[Software Carpentry]({{ site.swc_site }} ) was founded in 1998,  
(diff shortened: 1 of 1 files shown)


ChristinaLK:
@libcce there's an open PR that substantially modifies this page that I'd like to merge first, so this may require some fiddling after I merge it.  

@tracykteal (or @weaverbel?) can you confirm that this accurately reflects the current relationship b/n LC and the Carpentries as a whole?

weaverbel:
@libcce @ChristinaLK @tracykteal 
Re this in Chris's PR:
In May, 2018, The Carpentries Executive Council and the Library Carpentry Governance Group agreed to begin the process of aligning operations and Library Carpentry as a Lesson Organization. All  organizations will continue to maintain their own curricular materials.

@libcce I would reword it as follows:
In early 2018, The Carpentries Executive Council and the Library Carpentry Governance Group agreed to open discussions about bringing Library Carpentry under The Carpentries umbrella as a distinct lesson organization. Initially, while talks continue, The Carpentries will assist Library Carpentry with workshop requests and instructor training. Questions of governance, mission and vision, and other questions about aligning organizations will take place over the coming months.

lexnederbragt:
We have landed on the term 'Lesson Project', rather than organisation (as many non-profits have Projects as main activities, and to reflect that SWC and DC no longer are independent organisations). Could that be taken into account here too?

tracykteal:
Thanks for putting this in @libcce. We do want to include information about Library Carpentry in instructor training, so people know the current status and some background on the content and motivation of those lessons. I think @weaverbel language accurately represents the current state, and we can update to use 'Lesson Project' rather than 'Lesson Organization' as @lexnederbragt suggests. Given the merge conflicts with @ChristinaLK  PR, I would suggest closing this PR and putting in a new one with the updated suggested language.

libcce:
Thank you all for your suggestions. I will close this PR and incorporate your feedback into another PR.

PR #702, opened 2018-06-01
Clarifying point in expertise

### LINKED ISSUE #696: What does this sentence mean?
In episode 03 (expertise):

"Believing that something will be hard to learn is a self-fulfilling prophecy. This is why it’s important not to say that something is easy: if someone who has been told that tries it, and it doesn’t work, they are more likely to become discouraged."

I don't understand how the first sentence relates to the second sentence here. The first one seems to imply that we should tell learners that learning what we are teaching them is not hard. The second sentence, on the other hand, seems to suggest that we shouldn't tell them that it's easy, because then they will be discouraged. Am I misreading this as a contradiction? Is there any empirical evidence either way?

Thanks!

ChristinaLK:
Closes #696

### DIFF
--- a/_episodes/03-expertise.md
+++ b/_episodes/03-expertise.md
@@ -16,52 +16,52 @@ keypoints:
 ---
 
 We now discuss what distinguishes expertise
-from earlier stages of learning, how being an expert can make it 
-more difficult to teach novices, and some tools to help instructors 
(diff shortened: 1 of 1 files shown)


karenword:
Thanks @ChristinaLK !

PR #714, opened 2018-06-13
misc changes to the instructor guide

ChristinaLK:
Includes
* space for new material (slides + whatever else comes up)
* links at the top for trainers
* changing the minute cards link to a true template
* indenting the online teaching recommendations more sanely.  

I don't know if the internal links are going to work.

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -4,6 +4,10 @@ title: "Instructor Notes"
 permalink: /guide/
 ---
 
+Shortcuts to: 
+- Teaching Tips: [Online vs In Person](#-VI.-Differences-Among-Training-Types), [Online Teaching](#-VII.-Zoom-Manual-(Online-Trainings)), [General Recommendations](#IIX.-General-Recommendations)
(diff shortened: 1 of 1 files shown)


remram44:
Those numbered titles are very unusual :thinking: I see that they were already here, are they really required?

ChristinaLK:
I think @maneesha did the title formatting for the instructor's guide -- we could ask.  

There's enough headings in this document that I think numbers of some kind are useful.  If we removed them, I'd want to either split the doc into major sections or into separate pieces.

lexnederbragt:
Or use a table of Content, which I find can be quickly generated using https://github.com/ekalinin/github-markdown-toc (it parses the markdown and generates a Markdown TOC)

ChristinaLK:
@lexnederbragt I thought about a TOC but there's enough sections that I don't think it would be helpful, unless we did further grouping.

ChristinaLK:
@maneesha thoughts on this? I know you put together the current version of the notes.

ChristinaLK:
replaced by #773.

PR #749, opened 2018-09-07
make intro list match subheadings

brownsarahm:
minor fix, the list at the top didn't match the headings


ChristinaLK:
Thanks @brownsarahm!

PR #762, opened 2018-10-07
Letters to numbers for nested list rendering

NPalopoli:
Replaced letters by numbers (e.g. '(a)' to '1.') and accomodated spacing for appropriate display of nested list in section "Which of these questions assesses flaws in a student's mental model of a domain? You don't need to provide answers for these questions."

### DIFF
--- a/_episodes/01-welcome.md
+++ b/_episodes/01-welcome.md
@@ -147,22 +147,22 @@ motivate the activities and how you communicate with the attendees.
 >
 > *Which of these questions assesses flaws in a student's mental model of a domain? You don't need to provide answers for these questions.*
 >
->     1. I'm not sure what a mental model is.
->     2. "In Python, what is the expected output for the following statement: 1 + '2'"
-> 	(a) '12'  
-> 	(b) TypeError  
-> 	(c) ‘3’  
-> 	(d) 3  
->     3. "Rate your experience with the R programming language."
-> 	(a) never used it  
-> 	(b) beginner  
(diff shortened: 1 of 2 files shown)


ChristinaLK:
Will the sublist still render as letters?  These would be preferable to numbers, as most multiple choice questions are lettered, not numbered.

NPalopoli:
As far as I know, GitHub does not support lettered ordered lists. We could use HTML but I'd rather stick to markdown for simplicity and interoperability.

NPalopoli:
CSS styling of sublists (nested lists) as lettered lists, to overcome limitations of kramdown. Follows suggestions by @rgaiacs (#762)

ChristinaLK:
This looks good to me.  @rgaiacs, will this cause any problems w/ the upstream template?

rgaiacs:
> will this cause any problems w/ the upstream template?

You need to be sure that this only applies to lists inside the exercises. So you need something like

```
blockquote.challenge ol ol,
blockquote.challenge blockquote.solution ol ol {
}
```

And because we are using SaSS, we need to have a look if you can include it inside some other block. Sorry to not provide the full solution. Trying to mentor people. :-)

raynamharris:
Thank you @NPalopoli! I definitely agree that your proposed structure is a nice, easy fix for creating a multi-level list. However, we deleted this question entirely in PR https://github.com/carpentries/instructor-training/pull/771 because it was unnecessarily confusing to learners.

PR #764, opened 2018-10-08
Reorganize Overview to describe what Instructor training IS first, before

erjank:
saying what it is not.

### DIFF
--- a/_episodes/01-welcome.md
+++ b/_episodes/01-welcome.md
@@ -75,9 +75,10 @@ researchers to be more productive and do better research.
 
 This two-day training focuses on developing teaching skills that are broadly 
 useful across all of our lessons.
-These skills will be developed through reading pedagogical literature and through active
+These skills will be developed through active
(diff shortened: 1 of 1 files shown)


ChristinaLK:
This looks good to me.  @karenword ?

karenword:
I love the idea of this change! Two thoughts: 

- I suggest remove "reading pedagogical literature" since we don't really do that during the training. Maybe instead "These skills will be developed through discussion, practice, and feedback with fellow learners." ? Or something along those lines?
- I think the original intent of putting the negative first was to combat a very common misconception that this workshop teaches content (rooted in the widespread misconception that content proficiency is the most important prerequisite for effective instruction). To maintain this emphasis, I suggest bolding "not", and possibly the entire phrase "not be going over the workshop content"

raynamharris:
(The `close and comment` button is so close to the `comment` button.)

karenword:
Thanks @erjank and @raynamharris !

PR #774, opened 2018-10-19
adding mental models diagram

amyehodge:
I don't seem to have access to actually add the figure to the proper directory.


_episodes/02-practice-learning.md
```diff
@@ -92,6 +92,8 @@ does not have to be completely accurate in order to be useful:
 for example, the average driver's mental model of how a car works probably doesn't include
 most of the complexities that a mechanical engineer would be concerned with.
 
+![Mental Models](../fig/mental_models.jpg)
```
ChristinaLK: oops, this needs to be "jpeg"

amyehodge:
@ChristinaLK I think I've fixed this.

ChristinaLK:
Yep, looks good!

amyehodge:
@ChristinaLK sorry, I've been doing instructor training the last couple days and realized that the lesson contribution says images should be svg files. Should I get this to you in another format?

ChristinaLK:
If you can get a copy of the image as `svg`, that would be best.  Let me know if that's not possible.

amyehodge:
@ChristinaLK I made this in Powerpoint, and I don't think there is an svg option from there. I think I'd have to remake it in Illustrator or something and I do not have time to do that in the next week or two, but possibly later?

ChristinaLK:
Let's just keep the jpeg for now.  :)

PR #778, opened 2018-10-23
Redo day2 afternoon

ChristinaLK:
Comments welcome!  See a "live" version of these edits at: 
https://data-lessons.github.io/instructor-training/


kcranston:
Thanks for all of this work! I like the re-organization, especially the combination of the introduction exercises (one longer exercise instead of two short ones).

tracykteal:
Thanks @ChristinaLK I really like this. The section on introductions and updates to the exercises is really nice.

raynamharris:
So, if I'm correct, this is a summary of the proposed reorganization
screen shot 2018-10-23 at 1 38 33 pm

ChristinaLK:
omg @raynamharris thank you for the graphic. :heart_eyes:

anenadic:
Hi, nice one @ChristinaLK and a superb diagram by @raynamharris. Some of the links to the lessons in the 'live version' were broken so I could not see the new content. Structure wise, I like that Day 2 now starts with live coding as there is a build up expectation with learners to do so. It also separates it a bit from the second live coding session, and people will have more time to reflect on their potential mistakes and  the stuff we teach in between these two lessons that is meant to improve their live coding performance. My comment here is that this part *always* talkes longer than anticipated - last time I rushed through lessons study and people commented that they wanted more on lessons objectives. Also, the bit in lessons study about designing good challanges seems a bit repetitive as it is already covered in Day 1 (The Importance of Practice). So maybe ajust the times a bit - although this is tentative schedule so it does not matter much. I never finish Day 2 morning before 12:45-13:00 though :-(.

anenadic:
Not sure where creating workshop website fits now within the new schedule, last time I thought people felt a bit overwhelmed with it (and they were quite a technical bunch) and also a bit tired at the end of Day 2.

ChristinaLK:
@anenadic it would be right before the afternoon break in this model.

anenadic:
Ah, in "How we operate" section @ChristinaLK , got you. Good timing, people will be fresh(er) after the lunch break.

Issue #785, opened 2018-10-25
Data Carpentry Instructor Training


Valbat:
I think it would be good to add as a requirement or recommendation that the instructor give a quick summary about Data Carpentry's organization and mission and that the trainers are volunteers.

Valbat:
I would put the mission of the Data Carpentry organization up front and center. I would state the mission at the top of the page. In addition, one of the challenges for the instructor training is for the trainers to write a Data Carpentry intro for a class. Maybe you could pick a few from the instructor training classes and post them as examples for suggestions for opening for a Data Carpentry class.

raynamharris:
I took a stab at adding the mission statement for both orgs in the intro paragraph in #788. This may or may not be what you are looking for. I did not yet at a challenge exercise.

Valbat:
When a student takes the instructor training class, one of the exercises is to write an intro. for the data carpentry class. So, I thought the trainers could ask the students for their write ups of the introductions and if there are ones that the organization thinks rise to the top, they could post sample data carpentry introductions or templates of introductions that trainers give when starting a class that the organization would like to show case for new instructors.

Valbat:
My response above is to clarify what I meant about the challenge. It was an exercise/challenge during the instructor training to write up an introduction covering all the points highlighted during the class for how to start a data carpentry class.

raynamharris:
Thanks for the attempt at clarity, but I am even more confused. Is there or is there not an exercise about introducing Data Carpentry, perhaps in [this episode on introductions](https://github.com/carpentries/instructor-training/blob/gh-pages/_episodes/19-introductions.md)? 

Also, are suggestion that not only should an exercise/challenge exists but that it should be mandatory for checkout?

Valbat:
Hi Ray: If you have time today, please call me at 617-918-1674. I think it will be easier to explain over the phone. Thank you

karenword:
Hi @Valbat thank you for your persistence as we try to figure this out! I deleted your prior comment because I am concerned about having personal information like a phone number available in our public repository. While many things are more easily conveyed by telephone, we normally try to keep our communications about the repository here. 

My interpretation of what you're asking for is an example response for the activity in our Workshop Introductions lesson, which Rayna links to in her previous comment. I think that maybe you are using "Data Carpentry" to refer to any of our workshops, because it doesn't sound like you're looking for more background information (another kind of "introduction") about the Carpentries organizations or history. Does that sound accurate?

karenword:
[...]

PR #788, opened 2018-10-31
add DC and SWC mission

raynamharris:
Added DC and SWC missions as per #785 . This doesn't add a a challenge exercise as suggested, but it does provide a little more meat in the opening paragraph. 

One thing to consider is rewriting this whole episode. I think the venn diagram and some of the lists could be condensed into 1 table, for example.

### DIFF
--- a/_episodes/20-carpentries.md
+++ b/_episodes/20-carpentries.md
@@ -20,17 +20,21 @@ about how you can get involved.
 
 ## History
 
-[Software Carpentry]({{ site.swc_site }} ) was founded in 1998 and
-[Data Carpentry]({{ site.dc_site }}) was founded in 2013. You can learn more about the history
+[Software Carpentry]({{ site.swc_site }} ) was founded in 1998 with the mission of teaching lab skills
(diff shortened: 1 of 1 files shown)


raynamharris:
I know this didn't quite address #785 but is it a useful addition?

karenword:
Yes! However, I think we need to hold for inclusion of Library Carpentry in this summary.  Also I'm still not clear on terminology -- the Carpentries is supposed to be a project (of Community Initiatives) and not an organization itself, but can we still refer to the lesson-orgs as "organizations"?

raynamharris:
@karenword I added a sentence or two to clarify that is is a project (take from https://carpentries.org/about/)

i'm not sure what to do about adding LC. I can't find a blog post from which I can pull text....

raynamharris:
yikes!! I was trying to resolve a merge conflict (which turned out to be caused by a change in the file name) but I didn't mean to actually merge this PR! I was blindly following the command line instructors without thinking about the consequences. 

Not sure what to do next @ChristinaLK @karenword

karenword:
No worries @raynamharris -- I don't think there's any serious harm done here. We can leave it alone and update the language when we get more to work with on LC. There is a lot of work yet to be done with that merger, and writing these kinds of texts is on the to-do list.

raynamharris:
I find the post about LC! https://carpentries.org/blog/2018/11/welcoming-library-carpentry/ I'll add some text about it in the relevant sections

PR #790, opened 2018-10-31
Add direct link to demo instructions

karenword:
This is one item that Trainers may look for here on short notice, and it's not clear from the handbook link that these are located in the Trainer Guide.


raynamharris:
Thanks Karen! I modified the text a little so that the link to click said more than here. I remember in some previous issue or PR @arokem (I think) said more informative links are awesome.

PR #792, opened 2018-11-01
Create demos_rubric.md in extras folder

karenword:
Moves rubric to Extras where it should be properly renderable. Adds header.


raynamharris:
Did this not work as anticipated?

karenword:
works fine -- just intended to submit directly to data-lessons instead :)

Issue #784, opened 2018-10-25
mismatch between carpentries roles and and roles worksheet: need answer key


jt14den:
This came up in a instructor training we (@sheraaron) had today. There's a mismatch and confusion between the community roles <http://static.carpentries.org/community/> and the carpentries roles worksheet <https://carpentries.github.io/instructor-training/files/handouts/Carpentries_roles_worksheet_v2.pdf> that is used in [Episode 20](https://carpentries.github.io/instructor-training/20-carpentries/#participating-in-the-carpentries--whats-your-role). It would be helpful if the terminology matched and there was an answer key.

amyehodge:
@jt14den I made the worksheet based on the roles information that was in the curriculum at the time, but that has either changed or been removed from the curriculum and replaced with this link. Agreed, they no longer match up and need to. I will try to get this updated, but am not sure I'll make the deadline (which is when??).

jt14den:
@amyehodge There is no deadline. We wanted to flag the issue for future improvement.

ChristinaLK:
@amyehodge, yes no deadline for this change and you don't even have to do it yourself. ;)

amyehodge:
I probably have the original file still, so it should be easier for me than for someone else to create from scratch, so I'll get to it! Good thing to do when I'm procrastinating on those budget requests....

amyehodge:
[Carpentries_roles_worksheet_v3.docx](https://github.com/carpentries/instructor-training/files/2549801/Carpentries_roles_worksheet_v3.docx)
[Carpentries_roles_worksheet_v3.pdf](https://github.com/carpentries/instructor-training/files/2549802/Carpentries_roles_worksheet_v3.pdf)

I've updated this worksheet and attached it here, since I can't add the files to that folder directly. 

Answer key has been added via PR #794

PR #794, opened 2018-11-05
added solution to new matching exercise

amyehodge:
Note that this goes with the new handout and is not correct for the handout currently linked. This PR should be merged with the handout update to make sure these match. #784


amyehodge:
See issue #784

PR #810, opened 2018-11-13
Change 'coffee' to breaks to be more inclusive

lexnederbragt:
Not everyone drinks coffee, or even tea, and there may be cultural differences/sensitivities around this too. Too drastic?

### DIFF
--- a/_episodes/02-practice-learning.md
+++ b/_episodes/02-practice-learning.md
@@ -1,5 +1,6 @@
 ---
-title: "How Learning Works: The Importance of Practice"
+title: "Building Skill With Practice"
+block: "How Learning Works"
 teaching: 20
 exercises: 45
 questions:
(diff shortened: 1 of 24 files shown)


lexnederbragt:
My idea was to PR against https://github.com/data-lessons/instructor-training but Github chose otherwise. Shall I 'move' this PR over?

ChristinaLK:
@lexnederbragt submitting it here is fine!

raynamharris:
I agree with Christina about removing the word "short" since the time frame is specified as 15 min. 

Also, if we keep the word short, the capitalization needs to be fixed (as in "Morning Short break")

lexnederbragt:
Like this?

 (I recently learned that to undo changes to a file in a (PR) branch, you can use, while in the (PR) branch with unwanted changes, `git checkout origin/gh-pages filename`. Soooo useful! I used it here for undoing the changes in the `_includes` folder.)

ChristinaLK:
This looks good to me!  :D

PR #817, opened 2018-11-16
Add data carpentry github site

jcoliver:


karenword:
thanks @jcoliver ! Now should we link to these somewhere? :)

jcoliver:
It's referred to in episode 20, but there was no value for `site.dc_github`, so it ended up being self-referential. Discovered today while teaching. :D

Issue #766, opened 2018-10-11
choose your own adventure:  graphics for workshop layout


davis68:
I've a couple of graphics that I use in explaining how to structure SWC v. DC workshops.  I have found these helpful for students, but:
1.  Where would these make the most sense to introduce?
2.  What should be changed?

![outline-dc](https://user-images.githubusercontent.com/5383842/46833351-0a6fdd00-cd6e-11e8-8e9e-ba4a28d4071d.png)

![outline-swc](https://user-images.githubusercontent.com/5383842/46833352-0a6fdd00-cd6e-11e8-8068-0f247b3b180c.png)

k8hertweck:
I LOVE this! 

You may be able to simplify these by dropping Mercurial, MATLAB, and Make from the SWC lessons, as they don't appear to be considered part of the [core lessons](https://software-carpentry.org/lessons/) anymore. It might also be worth changing the order of appearance for scripting languages; I'm not sure whether this order is more commonly taught than bash, scripting language, then version control. Similarly, I've covered R/python prior to SQL in DC workshops.

It may also be nice to change the size of the circles to relate to amount of time spent; e.g., scripting languages often take two half days, so have a larger circle, whereas data organization and cleaning are both taught together in a morning (so smaller circles).

ChristinaLK:
These are GREAT.  :DDDDD Thanks @davis68!

Agree w/ @k8hertweck about dropping mecurial + Matlab.  I don't know how many SWC workshops cover SQL, either, but I'm happy to leave it in.  

For now, these could replace (or supplement) the venn diagram in: http://carpentries.github.io/instructor-training/20-carpentries/index.html

But we've also discussed doing more of an intro to what the Carpentries are at the start of the training, and if we make those changes, I'd put this there.

remram44:
I think this is a bit weird, because people (maybe CS people?) are bound to try and put meaning on those arrows (like dependency). While `Bash->Git` makes sense, `Git->Python` or `Python->SQL` is very surprising. Indeed, in the usual workshop schedule, SQL happens before Python is done.

karenword:
I also love this and think that it (or something very much like it) definitely needs to join the curriculum -- even where it is unclear, the diagram makes those points available for discussion in a way that the written text just can't. This offers a mental model of the relationships between workshop modules and it 'walks the walk' on concept mapping (even if it's mainly a flow chart). :) Thank you @davis68!

[...]

PR #822, opened 2018-11-26
Add workshop layout graphics

davis68:
This responds to and completes #766.

I've added workshop layout graphics for SWC/DC/LC to the Carpentries overview lesson.

On my screen, the images render at an appropriate width, but we can switch to  instead of Markdown inclusion if we need to.


davis68:
Currently the names are consistent with the base lesson names, or a direct abbreviation.  So one issue is that the naming schemes aren't consistent across Carpentries platforms.  Either that needs to be standardized (probably desirable, but the hard way) or we have to use alternate names here.

libcce:
@ChristinaLK I've asked the LC Governance Group what they think about the LC chart, but if I don't hear back from them, I think I can probably comment. I just want to check with them first.

karenword:
@davis68 what do you think about the prospects of making the colors more categorically consistent across diagrams? e.g. R/Python always being red, command line always being purple, data management type stuff always being blue etc? Also wondering whether these might be made a little bit smaller as we are worrying about the size of these -- it would be nice to be able to display them together if possible, or keep them separate but just less bulky on the page. The font size could probably be dropped without losing too much visibility, and some white space reduced?

If you don't have time for these changes that's fine -- we LOVE this idea and want to include it so we will find a way to make it work!

davis68:
I'd used and thought of them as slides, hence the size, but let me draft up something more compact and we'll see how we like it.

davis68:
Software Carpentry Workshop Elements
core lessons
	◐	The Unix Shell
	◐	Version Control with Git
	●	Programming with Python
	●	Plotting and Programming in Python
	●	Programming with R
	●	R for Reproducible Scientific Analysis
additional lessons
	◐	Automation and Make
	●	Programming with MATLAB
	◐	Using Databases and SQL
	◐	Version Control with Mercurial

How does this feel, suitably colored?

davis68:
image

Here's a more compact version with better organization in labels.

davis68:
Here is the most compact I can get them.  Preview at https://davis68.github.io/instructor-training/21-carpentries/index.html.

karenword:
@carpentries/core-team-workshop-admin @carpentries/core-team-community I am closing this PR because we are trying to point to canonical (external) resources for things like workshop information rather than maintaining content here. However, I think these graphics do a good job of visualizing a subject that our community-members sometimes struggle with. Maybe worth considering for related content in the website update?

PR #829, opened 2018-11-29
Add zoom host troubleshooting solution

lexnederbragt:
It has happened a few times that the instructor could not become host. This adds a solution to this problem.


ChristinaLK:
LGTM.

karenword:
@sheraaron what do you think about having "claim host" as the primary means of communication or including this in your email communications instead/as well?

raynamharris:
Thank you @lexnederbragt. We really appreciate this addition. Thanks for your patience with the merge.

PR #832, opened 2018-11-30
Update 05-memory.md w/method of loci

aurelg:
When introducing memory chunks, I think it could be useful to mention the powerful method of loci, and the link we know between spatial memory and cognition (FWIW, it could also be interesting to mention it with concepts maps).

(Could this PR be considered as part of the checkout process for my instructor's training online session Nov, 29-30th?)

### DIFF
--- a/_episodes/05-memory.md
+++ b/_episodes/05-memory.md
@@ -97,7 +97,13 @@ Storing words in groups reduces the number of effective
 items in your short-term memory, allowing you to keep
 more information in your head at once. This is one reason
 why helping our learners to see the connections among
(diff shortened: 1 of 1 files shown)


raynamharris:
Thank you for your contribution @aurelg. Yes, this counts as your contribution toward instructor training checkout! 

I discussed this with the other lesson maintainers and we are hesitant to incorporate this PR given that it adds another jargon term (method of loci), but we do really like the additional references and agree that this section of the lesson can be improved. Therefore, I'm are going to close this PR without merging but have opened Issue #835 for a discussion of ways this lesson can be enhanced.

PR #833, opened 2018-11-30
Added tip on randomising order of names

lexnederbragt:


ChristinaLK:
Next level exercise management! :)

PR #836, opened 2018-11-30
update timing

karenword:
Timing does not take the *nuts and bolts* exercise into account


karenword:
IMO it should be ok to add without subtracting — this day is currently scheduled to end at 16:45, so I think it’s ok to push this one for now as it still has us ending before 17:00. We can re-evaluate if things still seem to be running long.

Issue #842, opened 2018-12-03
Correction to history diagram


gvwilson:
`fig/SWCDChistory.png` refers to "Greg Wilson and Dr. Brent Gorda" - Brent isn't a PhD, so "Dr." should be removed. (Couldn't find source for the PNG, so was unable to repair and PR.)

ChristinaLK:
@raynamharris is this diagram yours?

raynamharris:
This was merged a while back. Closing now :)

Issue #843, opened 2018-12-03
Creation of Software Carpentry Foundation missing from history


gvwilson:
`fig/SWCDChistory.png` does not have a timeline entry for the creation of the Software Carpentry Foundation in 2014 (under the auspices of NumFOCUS).

raynamharris:
Merged a while back. Closing now :)

PR #844, opened 2018-12-03
update history figure

raynamharris:
remove Dr and Drs from names. removing one occurrence of Dr  was suggested in https://github.com/carpentries/instructor-training/issues/842 but then I decided to go ahead and remove all of them


raynamharris:
Added line to 2014 to indicate formation of SCF under numfocus as per https://github.com/carpentries/instructor-training/issues/843

raynamharris:
cc @gvgvwilson. I made it with illustrator. 

.ai files are a little large but I can upload it if you think that would be useful for the future

ChristinaLK:
This looks good to me.  
I don't think an .ai file would be useful but there might be another useful output format we could use.

PR #852, opened 2018-12-17
Updated language for revised discussion requirement

kariljordan:
The link to the Etherpad still needs to be updated, but I am not sure how to fix it. I see this:

`page.discussionpad`

But where is that linked from/to? The community discussions etherpad is https://pad.carpentries.org/community-discussions


ErinBecker:
Thanks @kariljordan for putting in this PR. Those links used to really confuse me to. The way they work is that, rather than putting in the full URL for a link, you create an alias for that link and then put it at the bottom of the document with the formatting `linkname: URL`. It makes it easier to update links in the future because you only need to update the alias at the bottom of the document, rather that replacing all instances of the link in the full document. Hope this helps!

PR #855, opened 2019-01-03
Include Library Carpentry in LICENSE

libcce:


rgaiacs:
Need to go into https://github.com/carpentries/styles.

libcce:
I believe this has happened already https://github.com/carpentries/styles/blob/gh-pages/LICENSE.md Is there more that needs to be changed?

rgaiacs:
> Is there more that needs to be changed?

@fmichonneau would normally make the updates to all the official lessons.

raynamharris:
As a follow up, I noticed that the LICENSE in styles was updated in commit https://github.com/carpentries/styles/commit/55c136ca83729a27ff80d3ea332b33ce813ae417#diff-37854d19817c792316d481f5beb93cc7.  

Should we close this PR without merging or go ahead and merge it even thought changes will be incorporated later?

raynamharris:
After getting some feedback from a fellow maintainer, we agree the best way forward it to mere here and later merge the styles repo. Thanks for your contributions @libcce!

PR #862, opened 2019-01-03
Add Library Carpentry lessons

libcce:


raynamharris:
Thanks @libcce for the PR. We recently moved all the template email messages to the handbook in https://github.com/carpentries/instructor-training/pull/866, so I'm going to close this without merging. I looked through the templates and all of them appear to be include LC. Here's the link to the handbook section for your reference https://docs.carpentries.org/topic_folders/instructor_training/index.html

PR #863, opened 2019-01-03
Add Library Carpentry language to icebreakers

libcce:


raynamharris:
This looks good to me, unless there is a standard somewhere that says we should use "Software Carpentry, Data Carpentry, and Libary Carpentry" instead of "Software, Data, and Libary Carpentry"...

PR #876, opened 2019-01-22
Fix for Dunning-Kruger effect link issue #875

anenadic:
Fix for issue #875


ChristinaLK:
looks good, thanks @anenadic!

Issue #891, opened 2019-02-25
Ep9 typo


gcapes:
At the end of this line is a strange character
>She’s written a very accessible popularized summary of her findings.

https://carpentries.github.io/instructor-training/09-mindset/index.html

raynamharris:
Thanks @gcapes!

PR #893, opened 2019-02-25
remove strange character

raynamharris:
hmm. there was a weird typo as pointed out by @gcapes in https://github.com/carpentries/instructor-training/issues/89/ ... maybe it was an emoji at one point, but not it was just a strange character or space. I removed it.


gcapes:
The issue is #891.
Thanks

PR #895, opened 2019-02-28
Add LC links to Submit a Small Contribution section

libcce:


_extras/checkout.md
```diff
@@ -58,12 +58,11 @@ Trainees must make a contribution to a lesson's content, exercises, or instructo
...
 
 Each lesson has a CONTRIBUTING.md file in its repository on GitHub. This file explains best practices for contributing
-to that lesson. You can access a lesson's GitHub repository by clicking on the GitHub kitty icon in the "Repository" column for that
-lesson on the lesson page. For Data Carpentry this is [www.datacarpentry.org/lessons/]({{ site.dc_site }}/lessons/). For Software Carpentry this is
-[www.software-carpentry.org/lessons/]({{ site.swc_site }}/lessons/). Please follow the guidelines in the appropriate CONTRIBUTING.md file when making lesson
+to that lesson. You can access a lesson's GitHub repository by clicking on the GitHub kitty icon in the "Repository" column for that lesson on the lesson page. For Data Carpentry this is [www.datacarpentry.org/lessons/]({{ site.dc_site }}/lessons/). For Software Carpentry this is
+[www.software-carpentry.org/lessons/]({{ site.swc_site }}/lessons/). For Library Carpentry this is [www.software-carpentry.org/lessons/]({{ site.lc_site }}/lessons/). Please follow the guidelines in the appropriate CONTRIBUTING.md file when making lesson
```
tracykteal: Is this URL correct for Library Carpentry? It has software-carpentry.org in it.

tracykteal:
Thanks @libcce. Added a comment checking if the URL is correct.

tracykteal:
Also, for Library Carpentry do you want to require all the lesson contributions come through github or also through email? For Data Carpentry we also allow them through email, because instructors aren't required to know github to teach DC.

_extras/checkout.md
libcce: So much for doing this quickly. Yes, I need to change this. Thanks for catching it!

libcce:
I opted for GitHub. Almost all of our contributions come via GitHub and my contact info is on the LC site if people want to contact me directly. Seems the best way? But I defer to you :)

tracykteal:
Given that knowing github also isn't a requirement for LC instructors, it would be a good idea to include email as an option for LC checkout. Since we already have that workflow available, it would work well and @sheraaron says it's working fine. So, then instead of changing the Software Carpentry sentence, you can change the Data Carpentry one to say "Contributions to Data Carpentry and Library Carpentry materials may be submitted through GitHub or by [email](mailto:{{ site.email }})."

kariljordan:
+1 to the e-mail option

libcce:
Via the site email, ok, that makes sense. Thanks! I made the update but also added bullets/pointers to the GitHub repos which I thought would be helpful since we mention that contributions can be made via GitHub.

tracykteal:
There's so many other things in our github organizations than just lessons, that directing people to that top level URL is unfortunately more confusing than helpful. The links in the following paragraph direct people to the lessons pages, which then you can use to get the github pages. Having those links clearer is a good idea though. Instead of having them in those sentences, we could pull them out as bullet points there.

libcce:
Thanks! Pulled them out as bullet points now.

tracykteal:
Thanks! Looks good.

PR #905, opened 2019-03-05
minor text addition re: problems & solutions

brownsarahm:
add note that the configuration problems and solutions page is linked from workshop pages, so instructors will know how to find it again


karenword:
Thanks @brownsarahm !

PR #908, opened 2019-03-13
Interactive and correctly oriented pdfs

Denubis:
Made PDFs interactive forms (in the most minimal sense)


raynamharris:
Thanks @Denubis

Issue #892, opened 2019-02-25
Ep9: Error framing not explained


gcapes:
https://carpentries.github.io/instructor-training/09-mindset/index.html

The concept of 'Error framing' is introduced with explanation (as far as I can tell).

raynamharris:
Thanks @gcapes. Forgive me for not getting to this soon. Please see #914 and let me know if that helps.

PR #914, opened 2019-03-20
Add error framing definition to mindset episode

raynamharris:
In issue https://github.com/carpentries/instructor-training/issues/892, @gcapes notes that error framing is not defined or explained. I tried to google a definition of error framing, but I could not find a website with a definition. There is a brief definition in [our glossary](https://carpentries.github.io/instructor-training/glossary/index.html), so I added that sentence into the lesson. 

Maybe someone has a reference to add? @ErinBecker @karenword ?


gcapes:
Great! Thanks

Issue #540, opened 2017-10-24
Website to test working memory


fmichonneau:
the site (http://cat.xula.edu/thinker/memory/working/serial) we use to test working memory requires Flash. As it's phasing out, it would be great to find a replacement that use another technology

ErinBecker:
Thanks @fmichonneau. That's good to know. If anyone has any suggestions, please add to this issue.

kariljordan:
[Here](http://opencoglab.org/memtest1/#interactive_test) is an interactive test from the Open Cognition Lab, but I can't tell if it requires flash.

fmichonneau:
the one mentioned by Kari doesn't require Flash

ErinBecker:
I taught with @symulation and @reedacartwright this week and we had some difficulties with the new memory game. 

- Not all of the symbols render on every operating system
- This activity takes a LOT longer than the previous version. Probably 3 minutes instead of 30 seconds. 
- The result isn't a number of items that they could remember, so isn't directly comparable to the 7 plus or minus 2 rule. It was a score on some normalized scale. 
- The activity doesn't work with the chunking exercise. Much harder to make stories/connections between symbols than words and also not enough time to repeat the exercise to illustrate chunking.
- There is an optional survey at the end before they get their results. You don't need to answer the questions, but there are about 5 pages of questions, so it's another few clicks. Also, the survey asks some awkward questions (seem to be randomized from a set, but the first time I took it there was a question about people from a particular location being stupid/lazy). Some of these questions could raise stereotype threat issues.

I highly recommend changing back to the original activity if possible, or testing out a new version to replace this one very soon.

ChristinaLK:
Should we just remove it for now?

raynamharris:
What if we replace the website with a verbal exercise that the instructor read aloud? See https://github.com/carpentries/instructor-training/issues/835.  The instructor could recite the following:

"I'm going to the market to buy 16 items: 
apples, blueberries, cherries, peaches, 
tea, coffee, beer, wine,
cinnamon, nutmeg, basil, oregano, 
doughnuts, muffins, croissants, bread.
What are the 16 items?
_pause_
Did anyone notice that these 16 items could be grouped in to four catagories: 
fruits, beverages, spices, carbs? Did grouping help you remember more words ?"

reedacartwright:
I think it is worth a shot. At the very least we could use the example temporarily until a better web-app is available.

tobyhodges:
[...]

PR #916, opened 2019-03-20
add note about flash to guide

raynamharris:
Given the discussion in https://github.com/carpentries/instructor-training/issues/540, this PR adds a note about flash not working for everyone. 

(I meant to include this with my change in https://github.com/carpentries/instructor-training/pull/915, but my github skills are not cooperating today).


karenword:
Yay, thanks @raynamharris !

Issue #911, opened 2019-03-15
Updating Carpentries Role Worksheet 


sheraaronhurt:
The [Participating in the Carpentries – What’s Your Role?](https://carpentries.github.io/instructor-training/files/handouts/Carpentries_roles_worksheet_v3.pdf) worksheet should now say Regional Coordinator and not Workshop Administrator.

amyehodge:
Sher -- since I made this, I'll put it on my list to update today or tomorrow. Though I may send it to you in email, because I seem to regularly do updates to this repo incorrectly. 🙄

amyehodge:
[Carpentries_roles_worksheet_v4.docx](https://github.com/carpentries/instructor-training/files/2994037/Carpentries_roles_worksheet_v4.docx)
[Carpentries_roles_worksheet_v4.pdf](https://github.com/carpentries/instructor-training/files/2994038/Carpentries_roles_worksheet_v4.pdf)

Here are the new files. I'll submit another ticket to for changing the link and the text for the answer key.

Issue #917, opened 2019-03-21
update link to worksheet and answer key


amyehodge:
For episode 21 https://github.com/carpentries/instructor-training/blob/gh-pages/_episodes/21-carpentries.md, need to update the link to v4 of the file. Also, the answer key should read "Regional Coordinators" instead of "Workshop Administrators."

The link can't be updated until issue #911 is fixed to update the file.

PR #921, opened 2019-03-29
update text and worksheet admin to coordinator

raynamharris:
update lesson text and the and worksheet, changing "Workshop Administrators" to "Regional Coordinator"

closes https://github.com/carpentries/instructor-training/issues/911 and https://github.com/carpentries/instructor-training/issues/917


raynamharris:
Arg. I thought I pull all the new changes so as to avoid this but I did not. Will close and try again.

PR #933, opened 2019-04-15
re-phrasing

aschuerch:
not being comfortable teaching is hardly the only motivation - sounded a bit dismissive to me, therefore I would like to propose this small change in phrasing


ChristinaLK:
This sounds okay to me. @karenword @raynamharris does one of you want to merge?

PR #937, opened 2019-04-23
Change wording of breaks

lexnederbragt:
See #810


karenword:
Thanks @lexnederbragt !

Issue #907, opened 2019-03-12
Confusing wording about "chosen lesson" for checkout


krismyu:
On the [checkout page describing the teaching demo](https://carpentries.github.io/instructor-training/checkout/#demo-lesson), it describes the "chosen lesson" as in the following:

> For your teaching demonstration, you will teach a short segment from your chosen lesson(s). We will pick the segment of the lesson(s) you are to teach on the day of the demonstration, so you must be prepared to teach any part of your chosen lesson(s)

It doesn't specify *who* is doing the choosing. At first I thought it meant that the people running the demo session would do the choosing and you would have to be responsible for teaching any lesson in any one of the carpentries.  It could be helpful to say somewhere that part of the procedure is to choose a lesson.

So one idea for a revision might be:

> Choose a lesson for your teaching demonstration. You will be asked to teach a short segment from your chosen lesson(s). We will pick the segment of the lesson(s) you are to teach on the day of the demonstration, so you must be prepared to teach any part of your chosen lesson(s)

ChristinaLK:
thanks @krismyu ! We're always looking to improve the description of the checkout process.

raynamharris:
Thanks @sstevens2 !

PR #942, opened 2019-05-02
Updating teaching demo lang

sstevens2:
This is a proposed fix to issue #907

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -201,7 +201,8 @@ please let us know in advance so that we can attempt to find someone to moderate
 We may not be able to accommodate all languages.
 
 For your teaching demonstration, you will prepare to teach a lesson from one of the
-Carpentries lesson programs. You will be asked to teach a short segment from your chosen lesson. 
+Carpentries lesson programs. You will be asked to teach a short segment from your chosen lesson from this 
+[list of suggested episodes](https://carpentries.github.io/instructor-training/demo_lessons/index.html). 
 The host of the session will pick a segment of the lesson for you to teach on the day of the 
 demonstration, so you must be prepared to teach any part of your chosen lesson.
 
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -200,10 +200,11 @@ If you want to do your demonstration in a language other than English,
 please let us know in advance so that we can attempt to find someone to moderate.
 We may not be able to accommodate all languages.
 
-For your teaching demonstration, you will teach a short segment from your chosen lesson(s).
-We will pick the segment of the lesson(s) you are to teach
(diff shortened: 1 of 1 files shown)


karenword:
The only downside to linking to the table is that explicitly identifying episodes undermines the intent to have learners prepare a full lesson rather than (in some cases) a single episode which is the only one that happens to work well for live coding. It also gives learners the option of selecting lessons based on the number of episodes that they may have to prepare. 

The original version of this table was intended to address this need and was missing episode identification specifically for this reason -- I failed to notice that this was being lost when @maneesha ported the full table here from the handbook. I'm not sure how (or whether) to address this now. Thoughts?

maneesha:
We do tell trainees that the demo session leader will choose any episode from within their chosen lesson.   But trainees should also feel assured that the chosen lesson will be suitable for a live coding demo -- that we won't choose something that has no live coding, or something that has dependencies. 
  I've had trainees choose a lesson that had no suitable episodes within it, and I don't want people to be in that situation.  We do want learners to prepare a full lesson, and we can emphasize that during the training.  

I think it's OK to have this here.

sstevens2:
I think I'm alright with the option they may 'cheat' and not prepare the whole lesson.  I think trainee's typically choose their lesson based on what they are most comfortable with or planning to teach first rather than which will be the easiest to prepare.

sstevens2:
I've also been reminding folks who sign up for my teaching demo that they need to prepare the whole lesson a day or so ahead of time.

brownsarahm:
I thought about proposing similar changes after Lex's question yesterday and found this.  

Maybe we should also state that the lesson must include live coding or equivalent, both here and in the checkout episode?

ErinBecker:
Hi folks - I haven't been following this issue closely, but do want to note that for some lessons, we allow trainees to use episodes that don't have live coding, per se - so long as they have something that the trainee has to work through to demo (not just talking). This includes the Spreadsheets lessons (DC Ecology, Genomics, Social Sciences, LC) and OpenRefine lessons (same curricula). It may also include other lessons I'm not aware of. So I would be cautious about saying that the episode has to have "live coding".

brownsarahm:
@ErinBecker of course! that's what I meant by "or equivalent"  but that's probably not clear word choice. Maybe:

```
Your selected lesson must include episodes with either live coding or software demonstration, such as those included on this [list of suggested episodes](https://carpentries.github.io/instructor-training/demo_lessons/index.html).  You get to choose the lesson (the first column of the table in advance) and the trainer running your demo session will choose the specific episode for you during the session. 
```

ErinBecker:
Thanks @brownsarahm, that language looks good to me!

PR #944, opened 2019-05-03
Update demo table page with lessons from handbook table

maneesha:
Updating the demo page with the lessons that are listed in the handbook.  Table styles in the curriculum are *much* easier to read than table styles in the handbook.  Would be great to have someone who can build this locally review before merging.

Once this is merged I will remove what is in the handbook and link back to here.


maneesha:
Note I can't figure out how to make the columns the same width for all the lesson tables.  This may require formatting the tables as html rather than md.

raynamharris:
Thanks @maneesha! A previewed it locally and it looks good there. I do see what you mean about columns not being equal, which would be preferable. Let me look into html tables real fast. 

Screen Shot 2019-05-03 at 9 53 40 AM

raynamharris:
Maneesha, I didn't find a quick solution to fixed columns width, so instead, I modified the tables to make the content and style of each table more similar. Now it looks like this

Screen Shot 2019-05-03 at 10 45 55 AM

Screen Shot 2019-05-03 at 10 46 04 AM

ChristinaLK:
this looks nice!

maneesha:
Thanks @raynamharris  and @ChristinaLK  for the review.  Merging now.

PR #945, opened 2019-05-03
Clarified statement about which contributions are welcome

dhuppenkothen:
Having just done the 2-day instructor training, one thing I found a little bit confusing when reading through the checkout procedure was the statement about what (not) to contribute, especially since the instructor trainers had just told us they were always looking for new exercises for the existing materials. 

I thought it might be worth clarifying that while there's not really space to introduce new concepts, new exercises on *existing* materials are welcome (assuming I understood this correctly?). Otherwise, there might be other people like me who misunderstand and don't contribute their ideas for exercises. 

As an aside, is it too meta to submit a PR on the checkout procedures lesson as part of the checkout procedure? :)

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -100,7 +100,10 @@ Please note that:
     {: .source}
 
 5.  We do *not* want exercises or other material that introduce new concepts:
-    our lessons are already full to bursting.
+    our lessons are already full to bursting. However, new exercises for concepts that
+    already exist in the materials are very welcome. They allow instructors more 
+    flexibility to choose from when teaching the material to a class with diverse 
+    skills.
 
 6.  We are happy to accept contributions of new diagrams for any of our lessons.
     Please use SVG (scalable vector graphics),


ChristinaLK:
This sounds right to me! Thanks for the clarification @dhuppenkothen.  :)

@karenword @raynamharris can you confirm that this is indeed what is meant by this section and then merge if so?

karenword:
@ErinBecker can you determine whether this PR reflects the views of the maintainer community and, if not, suggest an alternative clarification? I agree that this section creates mixed messages.

ErinBecker:
Thanks @karenword for looping me in and @dhuppenkothen for this contribution. (It's definitely not too meta to submit a PR to the checkout instructions as part of your checkout!)

This is perfect timing - we have our monthly Maintainer meeting tomorrow. I've added this to the agenda and will ask folks to add their opinions to this PR thread. As a caveat, it's highly unlikely that there is a single "view of the maintainer community" so feedback will probably be mixed. 

My two cents for a slight rewording: 

> our lessons are already full to bursting. However, new exercises for concepts that
    already exist in the materials are very welcome. They allow instructors more 
    flexibility to choose from when teaching the material to a class with diverse 
    skills. When submitting your PR, note which concepts from the lesson your exercise tests
    and (if applicable) which of the current exercises it might replace or supplement.

k8hertweck:
I would be amenable to reframing the recommendations for submissions to redirect folks from exercises and lesson content altogether. It seems (for most lessons) that commenting on existing issues and PRs would be most useful. Alternatively, filing an issue or submitting a PR relevant to the instructor notes would provide the most improvement to the broader instructor community.

raynamharris:
Thanks @dhuppenkothen. This is an excellent point. You inspired me to do a little reformatting of this section. Do you think this helps or make things worse?

dhuppenkothen:
I like the new rewording a lot. I think that does make it clearer, and it also sounds a bit more encouraging and positive than the old text did. :)

gcapes:
> commenting on existing issues and PRs 

I agree with @k8hertweck - this is a useful way to contribute to lessons

PR #950, opened 2019-05-06
Revised avenue for reaching the Code of Conduct Committee

kariljordan:
The Code of Conduct Committee now collects incident reports through a secure form. Updating the instructor training to reflect this change. I'd be grateful if either of the maintainers reviewed and merged this change so that it is consistent with what is in the handbook.


karenword:
Yay thank you @kariljordan! This looks like it solves the most immediate problems we need to address. I suggest we merge this now to be current with the Handbook.

In the long run, I'd like to suggest adding some text that encourages people to default to reporting in cases where they are unsure whether something constitutes a CoC violation or not. But this could be added as part of the broader re-work in how we teach this content, along with some guidelines on hosts & Title IX etc.

PR #951, opened 2019-05-11
add  Spanish lessons to suggested demo episodes

raynamharris:
This creates a table of all the Spanish episodes. Copied from the English lessons above and updated with the titles and hyperlinks to the Spanish lessons


ChristinaLK:
I'm going to assume these are all correct because I'm not familiar with the lessons. :) 

Formatting looks good to me.

PR #956, opened 2019-05-14
adding note about day 2

### LINKED ISSUE #850: Wondering about small shift to lesson order...
After teaching a few of these both in-person and online I've noticed a spontaneous schedule switch that repeatedly happens that it might just be worth codifying directly into the suggested lesson order.

What happens is that we send participants home at the end of day one with a request to look up more about the Carpentries workshop types and then we ask them what questions they have at the start of day two.  While the intent is that these questions will be put in the etherpad and then taken up later in one way or another this doesn't happen or at least only happens in part.  The result is either a lot of saying, "We'll cover that this afternoon" or pulling parts of the relevant afternoon sessions into the morning.  Participants thus either end up a little frustrated (because they'd really like an answer now) or the schedule ends up being a little fractured.

Is this the experience of others?

Would there be interest in taking "Checkout Process" and "How We Operate", dropping them right after "Welcome" on day two, and then shifting everything else down accordingly?

ChristinaLK:
closes #850

### DIFF
--- a/_extras/guide.md
+++ b/_extras/guide.md
@@ -204,7 +204,7 @@ As written, this can run long.  Suggestions:
 ### Day 2 general
 
 * The official curriculum has been re-ordered over time and individual trainers have
-experimented with the order of the day 2 episodes. One example/motivation
+experimented with the order of the day 2 episodes. One example/motivation:
   * From JS: "What happens is that we send participants home at the end of day one with a
   request to look up more about the Carpentries workshop types and then we ask them what
   questions they have at the start of day two. While the intent is that these questions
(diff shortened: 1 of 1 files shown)


karenword:
Thanks @ChristinaLK ! Is it worth adding a secondary solution? When I've done this we've had both trainers working together to answer questions in the etherpad during live coding practice, so everyone gets their questions answered right away, often with extra detail & links, and we only have to take time to come back to them with a mention here and there.

ChristinaLK:
I'll try to add that in.

ChristinaLK:
@karenword have added, take a look and let me know what you think.

PR #968, opened 2019-05-20
Removed language about "final stages"

kariljordan:
We say they do not need to complete the checkout process in any particular order, so I removed "final stages" from the language of the rubric.


karenword:
Thanks @kariljordan ! Great catch!

PR #965, opened 2019-05-19
smoothing #2: classroom environment material.

ChristinaLK:

PR #970, opened 2019-05-21
Remove self-fullfilling prophecy about scientist not wanting to teach.

BenjaminSchwessinger:


BenjaminSchwessinger:
Ups I meant to say not wanting to learn programming.

karenword:
Thank you for addressing this @BenjaminSchwessinger ! I am not sure this fully solves the problem, though. What do you think about an alternative fix that avoids generalizations about scientists entirely? Maybe it could be more about programming and less about scientists, e.g. "Programming is inherently interesting once it is familiar, but for those who encounter it as an obstacle to progress in their work, curiosity doesn't always come naturally at first." ?

Denubis:
Perhaps more specificity is useful? "The first time many people encounter a useful program which needs some customisation, the effort or expense seldom repays the effort as we're leaning the patterns of the program as well as how to get the result we want. Once we know how to think in the fashion the program wants us to think, the effort invested is well repaid -- but it's hard to see that when we're just starting out"

BenjaminSchwessinger:
Late to this one. Yes I agree it could get away with any generalization e.g. 
"This presents us with a problem because believing that something will be hard to learn often becomes a self-fulfilling prophecy."

What you think?

ChristinaLK:
@BenjaminSchwessinger I like the last change you proposed. Can you add it in and we can see how the whole thing reads?

Also slight typo: imitating -> intimidating

ChristinaLK:
@BenjaminSchwessinger I have an open PR on this episode and am going to cherry-pick your changes into that (I hope).

ChristinaLK:
included in #965

PR #974, opened 2019-05-26
Fix links

fmichonneau:
in preparation for the lesson release, we are checking the lessons for broken links. This PR addresses the ones that were found.

As a reminder: 
- using `{% link path/to/page_in_repo.md %}` is the [preferred way to make internal links](https://jekyllrb.com/docs/liquid/tags/#links) in Jekyll. The `{% link %}` command figures everything out and the links are guaranteed to work (otherwise there will be a built error). I fixed the broken ones here but there are others that use the "old" way that uses `page.root`.
- when possible, use relative links (using the link command as above) rather than pointing to the rendered source (e.g., use `[motivation episode]({% link _episodes/08-motivation.md %})` and not `[motivation episode](https://carpentries.github.io/instructor-training/08-motivation/index.html)`). With the latter, the link in the rendered site will not work when used offline.


raynamharris:
Thanks! @fmichonneau

PR #976, opened 2019-05-26
rearranging teaching feedback section

ChristinaLK:
these are sort of (?) major changes, so feel free to veto.


ChristinaLK:
@karenword I agree wholeheartedly about the fixed mindset thing. I took it out because it interrupted the flow of ideas as I was editing them, but I definitely think it should be there. I'll think on this and see where I can work it back in.

ChristinaLK:
I made these changes on my personal computer, which is why I hadn't pushed them! @karenword this is ready for another review. Let me know what you think. :)

Issue #960, opened 2019-05-15
"Wrap up" episode needs learning objectives


raynamharris:
As part of the lesson release (#934) all episodes need learning objectives. The checkout episodes where they give feedback about the workshop does not have learning objectives. https://carpentries.github.io/instructor-training/25-wrap-up/index.html

Any suggestions?

ChristinaLK:
Maybe make the question into the objective and change the question to "What can we improve about these training?'

chrispunjoku:
Excellent observation @raynamharris . Yes @ChristinaLK Better still, we can have such statement as:
At the end of this episode, you (participants) will have provided feedback that will help in improving the contents and delivery of this training. 
or
1) Give one up, one down feedback;
2) Fill in a few sticky notes;
3) Complete a post-workshop survey.

How do we see any of these?

PR #982, opened 2019-06-01
Update 25-wrap-up.md

kekoziar:
This episode was missing objectives.  I created active learning objectives based on my experience of the last session during instructor training.  This addresses issue #960


ChristinaLK:
This looks good to me! Thanks @kekoziar

PR #991, opened 2019-06-06
Add etherpad info to the lesson 

brownsarahm:
I added the pad of pads as a callout in the checkout section, inspired by @DanielleQuinn's choice to introduce it there in our workshop this week. 

and added how to create an etherpad to workshop setup.


ChristinaLK:
@brownsarahm this is awesome!

I'd maybe put the callout about the etherpad  into the Carpentries how we operate section (maybe in the section about joining the Carpentries community?) but I'd rather merge this as is if you don't have time to make that change. Let me know. :)

karenword:
I kind of like having it in Checkout actually -- can we add it somewhere in both pages? This is the page that I think people will return to when they are trying to figure out what to do and where to go, so having the call-out here I think will be most effective at reminding them that this resource exists.

ChristinaLK:
Let's just merge then, and we can add it to the carpentries how-we-operate at our leisure. :)

ChristinaLK:
thanks again @brownsarahm !

Issue #999, opened 2019-06-21
Bug in Wrap-Up-Doc PDF 


k8hertweck:
One of my learners (@alliblk) noted that the [Wrap-Up-Doc PDF](https://carpentries.github.io/instructor-training/files/handouts/Wrap-Up-doc.pdf) doesn't function for adding text; she will add more information below.

sstevens2:
We also had in issue with this handout in my training today, if you write something in one box and then click on another box, it copies what you wrote into a bunch of the other boxes. 
Maybe a form where they could fill it out and then input their email address to have it send it to them might be a good solution for online trainings?

pennyjw:
This is happening because when the form was constructed the fields were copied to make multiple blanks, so there are 5 fields called 'Text8', and text entered in one of them will be represented in them all. This can be fixed by renaming the fields so that each has a unique name in the Form fields view. These names ideally are semantic, representing what the expected content of the field is. There are also several other accessibility problems in the form.

ChristinaLK:
Changes / updates welcome! I made the doc in a hurry for the fall lesson release. If someone knows more about creating accessible pdfs, that would be awesome.

ChristinaLK:
@d-diaz updated version is merged. :)

PR #1001, opened 2019-06-21
Updated PDF templates to fix replicating text bug

d-diaz:
Also adds multi-line option for filling in text boxes.

Related to #999.


ChristinaLK:
Looks great, thanks @d-diaz !!

Issue #1007, opened 2019-07-01
broken links inside lessons


aschuerch:
(Some of?) the links that link to another lesson do not work, i.e. in https://carpentries.github.io/instructor-training/14-live/index.html the link to error framing strategies links to https://carpentries.github.io/09-mindset/index.html#errors-are-essential-to-learning instead of https://carpentries.github.io/instructor-training/09-mindset/index.html#errors-are-essential-to-learning. I will attempt to create a pull request to fix this

maneesha:
Another broken one - the icebreakers link (https://carpentries.github.io/icebreakers/index.html) from here (https://carpentries.github.io/instructor-training/01-welcome/index.html)

maneesha:
`_config.yml` sets the `training_site` variable to `"https://carpentries.github.io/instructor-training"` but when that's rendered by using `[icebreaker question]({{training_site}}/icebreakers/index.html)` the repo name seems to get stripped.  I don't know why this is happening but can look into this more (although probably not until after I teach IT this week).

maneesha:
Links that are defined using `{{ page.root }}` seem to work.

ChristinaLK:
@fmichonneau said that we shouldn't use `page.root`. :(

maneesha:
Linked PR is: https://github.com/carpentries/instructor-training/pull/1008

PR #1008, opened 2019-07-01
fixing links in episode 14

aschuerch:
I think this fixes issue #1007. Untested, though

### DIFF
--- a/_episodes/14-live.md
+++ b/_episodes/14-live.md
@@ -69,7 +69,7 @@ We refer to the practice of having the instructor live code and the learners cod
 >> have questions not directly addressed in the text of the lesson.
 >> * It can be hard for learners to listen and type at the same time, due to the  
 >> *split-attention effect* we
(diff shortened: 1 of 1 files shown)


ChristinaLK:
@fmichonneau I'm still not 100% familiar w/ this jekyll shortcut -- is this the correct format?

fmichonneau:
thanks for the report @aschuerch! 

These link issues are tricky given our setup and constraints. In general, we shouldn't use `page.root` in our lessons as the value of the variable will sometimes generate absolute paths that don't work when rendered locally.

It's difficult to debug this locally as they are many moving parts. As a first step, @carpentries/instructor-training-maintainers would you be OK with upgrading to the latest version of the template (I'll open a PR) as it attempts to fix some of these link issues. If that doesn't fix the broken links reported by Anita at least it will make it easier to find a fix. Thanks!

ChristinaLK:
Waiting on #1009

karenword:
This was fixed in #1022 . Thanks @aschuerch !

PR #1009, opened 2019-07-03
Upgrade to latest style

fmichonneau:


ChristinaLK:
@fmichonneau I don't even know what I should be looking for in this PR to check things. I assume I can probably just blindly merge? ;)

fmichonneau:
yes, I tested locally and it should be OK to merge. I'll look into #1008 when it's merged. Thanks!

PR #1010, opened 2019-07-05
Added etherpad exercise exporter.

Denubis:
Used the aio.md as inspiration for an exercise outputter to paste the current outline (h1, h2, h3), exercises, discussions, and callouts into a format suitable for copying into an etherpad.

Also added a line break in the testamonial opening so it appears correctly on two lines.


ChristinaLK:
@Denubis thanks for this! Can I ask how it's suppose to work / how you're supposed to get the text to paste into the etherpad?

Denubis:
Sure thing. Here's how I used it in my recent ttt.

I opened: https://denubis.github.io/instructor-training/ (as my forked site) and went to extras, Etherpad Template.

That uses a all-in-one technique similar to the all in one beta to produce:
Screenshot from 2019-07-24 06-54-21

I then have the page run some jquery to eliminate all sections that I felt shouldn't be on the etherpad. (This is very flexible and I anticipate discussion here)

I then hit select all, copy. Went to the etherpad and hit select all, paste. (While I could do javascript magic, javascript magic involving the clipboard isn't particularly OS independent and so I went for the simpler route.) This is intended as a drop-in replacement for the current etherpad template.

_extras/etherpad.md
```diff
@@ -0,0 +1,116 @@
...
+  $(this).text("Callout: "+oldtext).before("<br/><br/>").wrap("<b>").wrap("<i>");
+  
+});
+$("blockquote.discussion h2").each(function(){
+  var oldtext = $(this).text();
+  $(this).text("Discussion: "+oldtext).before("<br/><br/>").wrap("<b>").wrap("<i>");
+  
+});
+
+    console.log("hi");
```
maxim-belkin: `console.log("hi, there")` :)

_extras/etherpad.md
```diff
@@ -0,0 +1,116 @@
...
+
+
+
+
+{% comment %}
+Create anchor for each one of the episodes.
+{% endcomment %}
+
+{% for episode in site.episodes %}
+<h1>Episode: {{episode.title}} <a href="{{site.url}}{{site.baseurl}}{{episode.url}}">{{site.url}}{{site.baseurl}}{{episode.url}}</a></h1>
```
maxim-belkin: This would not render locally. Consider using `{{ relative_root_path }}`. Even better -- have a look at `_includes/all_keypoints.html`

ChristinaLK:
@Denubis cool! I realized after the fact that I couldn't see the template on the main repo page because this hadn't been merged yet. D'oh!

You can copy and paste the rendered text and it transfers okay to the etherpad?

maxim-belkin:
I was asked to have a look at this PR because I previously contributed to `aio.md`.

First, I would suggest submitting unrelated changes in different pull requests: the change to `01-welcome.md` is not related to the main change -- new file `etherpad.md`. While it might seem like a minor thing, it does help troubleshoot issues when they're encountered. I'm sure maintainers of this lesson will be happy to help with splitting the PR into two.

Second, there is some serious jQuery "magic" here... in a sense that without any comments it will be time-consuming to understand what and, most importantly, why each line is doing even 3 months from now. Of course it is kind of straightforward, but comments would be helpful.

Third, I would also suggest stripping off "output" cells.

Finally, some links will not work correctly when rendered locally. I suggest either fixing the links or detecting the case when the site is rendered locally and not adding links to episodes in such cases.  I made a comment about this above, see https://github.com/carpentries/instructor-training/pull/1010#pullrequestreview-266160745.

Denubis:
@ChristinaLK This is what I used for my recent ttt: https://pad.carpentries.org/2019-07-18-ttt-macquarie and to render my lecture notes for it.

Hi @maxim-belkin I'll work on this... probably this weekend. Agreed on comments. And I'll see what I can do to remove wrapping blockquotes, as that's a relic of some of the hacks I did to get things to vanish.

A point of confusion:
When I render the site locally, via `bundle exec jekyll s` I get: 
Screenshot from 2019-07-25 07-52-27

Which matches my intention of an absolute path to be copied to other sites. Can you provide a repro of the scenario you're talking about so I can better test? I will investigate what you have suggested in the meantime.

I'll live the splitting of the PR to other folk. Happy to just remove the insertion of the linebreak.

_extras/etherpad.md
Denubis: Question: I don't see relative_root_path in all keypoints, though this suggestion is really good because it reminded me to add the markdownify filter.

Looking at this, I *think* I'm getting what you're saying. You're suggesting that baseurl isn't operating the way we need it to do and we should swap baseurl for relative-root-path, rather than site.url *and* baseurl. Since I've started editing this, I'm going to throw a commit up for your review.

maxim-belkin: > I don't see relative_root_path in all keypoints

This means that this lesson hasn't been synced up with https://github.com/carpentries/styles in a while. `relative_root_path` is defined in here: https://github.com/carpentries/styles/blob/gh-pages/_includes/base_path.html  and all_keypoints are here: https://github.com/carpentries/styles/blob/gh-pages/_includes/all_keypoints.html

> You're suggesting that baseurl isn't operating the way we need it to do and we should swap baseurl for relative-root-path, rather than site.url and baseurl.

If I `jekyll build` the site, the links will be `/01-episode-name/index.html` and so on. With `relative_root_path` they will be correct

maxim-belkin: Actually, there is `all_keypoinds.html` in this repo, see https://github.com/carpentries/instructor-training/blob/gh-pages/_includes/all_keypoints.html

Denubis: @maxim-belkin Yes, the last time I pulled my repo from this one was before the June push. Thank you for the updated reference. I've changed {{site.baseurl}} to be {{relative_root_path}} as per your request. However, we still require {{site.url}} otherwise the pad.carpentry.org links will point to other etherpads instead of lessons. My links are absolute links such that these notes resolve to the original lesson regardless of where they are pasted. Sorry, I feel like I'm not communicating well this morning.

maxim-belkin: > I've changed {{site.baseurl}} to be {{relative_root_path}} as per your request. However, we still require {{site.url}} otherwise the pad.carpentry.org links will point to other etherpads instead of lessons.

I think you're right. I was thinking in terms of links within a website.

Denubis: All good. The ghost of wittgeinstein is laughing at our language games. :) In terms of relative_root_path v. baseurl, do you have a sense is which is more appropriate to use with site.url?

twitwi:
This seems to be an interesting feature, and it probably should not be only in the instructor-training repo but applied to all lessons.

I was thinking that copy/pasting all the content at once might be confusing (or spoiling) for the learners, and imagined a single-page approach that would load both the lesson content and an etherpad (to be able to control it directly) and allow easy one-click copy/paste of an exercise into the etherpad.

I'm reporting here some experiments that I just did around this.
It is verbose and most of it is probably useless (e.g. trying to have everything in a single page), but it is also a way of documenting my exploration.
See below for the instructions for a _demo_.

- the prototype targeted the injection of individual callout/exercices, one-by-one, when needed,
- browser security prevented the actual interaction with the etherpad if loaded in a frame (or similar tricks),
- ... one solution has been to inject (via the browser console, see below) some "plugin" for an etherpad
- I could access the etherpad chat easily but not really the etherpad editor for the injection of HTML content,
- pasting programaticallly did not work either,
- ... so the best I could do is automatic-copy and drag/drop
- ... but no automatic-paste... which makes the single-page approach less useful, unless we want to automatically post messages in the etherpad chat...
- in _conclusion_, it is probably better to have a generated page (as in the PR) unless there is a good way to interact with the editor

**Want a demo?** of this prototype, then:

- open a new test pad, e.g., https://pad.carpentries.org/test-issue1010
- open the browser console (Ctrl+Shift+K ?)
- open https://dl.heeere.com/etherpad-addon.js and copy/paste its first two lines in the console
  - these lines will sefl-load the file in the current page
  - your browser might ask you to type "allow pasting" if you're not usually doing web development
- you should get an interface that allows
  - to select which lesson to load (it directly loads the aio file and each episode)
  - to click on an exercise/callout to copy its content to the clipboard
  - to drag and drop an exercise/callout into the etherpad
  - to fold the added controls with ⇵
  - to reload the plugin with 🗘 (for development or when one type a new lesson name manually)

twitwi:
I actually found a way to access the editor (while writing the previous comment) so I'm updating the comment and the demo, we can now click-to-insert in the etherpad.

twitwi:
This can also be setup as a userscript (e.g. with tampermonkey http://www.tampermonkey.net/ ):

~~~
// ==UserScript==
// @name         Etherpad Carpentries
// @namespace    http://tampermonkey.net/
// @version      0.1
// @description  try to take over the world!
// @author       You
// @match        https://pad.carpentries.org/*
// @grant        none
// ==/UserScript==

(function() {
    'use strict';

    var injectPlugin = () => {
        var s = document.createElement('script')
        s.setAttribute('type', 'text/javascript')
        s.setAttribute('src', 'https://dl.heeere.com/etherpad-addon.js')
        s.setAttribute('class', 'pluggedIn')
        var head = document.getElementsByTagName('head')[0]
        head.insertBefore(s, head.firstChild)
    }
    injectPlugin()
})();
~~~
(this version loads the full fill everytime, an alternative is to put the complete file as a script)

maxim-belkin:
@twitwi, given that carpentries host etherpads, can you add a combobox(-es) and a button to the navbar so instructors could select the lesson and the episode and then click "Copy" button to do what your script does?

twitwi:
If we can/want/accept to modify the hosted etherpads, we can integrate something like this yes (maybe hide it by default so that non-instructors are not tempted to use it)

If the goal is to copy all the content (to replace or append to the etherpad), then it should be pretty easy. If there are needs for a more advanced interface (like I tried to do, with copy of the exercises one-by-one etc), we should try to specify it better.

In any case there is a decision to make: 
- do we need an actual new "extra" (so that the etherpad modification is minimal and rarely updated)
- or do we use directly aio.html (or aio/index.html) as my script does (then the formatting logic is in the etherpad instance)

twitwi:
(the modification to the etherpad instance could also just add something that injects the code from somewhere else)

Denubis:
@twitwi I find myself somewhat... hesitant to include instructions that have arbitrary javascript running on folks computers (this is the point where Karen would note that I'm suffering from expert blind spot, I think), at least in this "minimum viable product" pull request. Do you have a sense of how we can export data from the page such that instead of copying and pasting, we can use etherpads import "html/whatever" patterns? My intention is absolutely to migrate this to the general lesson template if it proves useful, but since I used it for ttt, this is where I'm starting.

@maxim-belkin have I addressed your concerns with my most recent commit?

twitwi:
@Denubis you're totally right with executing arbitrary javascript and I vote for going on with the current version of the PR and (potentially) thinking about a more advanced thing at another time.

Overall: I think it can be ok to merge this, then integrate it for all lessons and later, consider whether something more (improvements to etherpad for our use cases) could be interesting.

maxim-belkin:
> @maxim-belkin have I addressed your concerns with my most recent commit?

Looking at this right now...

maxim-belkin:
Overall: I think it will be very useful for instructors.
Specific comments:

* Episode Questions and Episode Comments appear right underneath "Episode: Title", so there is no need to say "Episode" again. Moreover, I think that "Episode" can be removed from everywhere. It's not important how we call it (episode, lesson, theme, act, _etc._).
* Lists (of objectives and questions) are spaced with empty items when copied to etherpad:

    > Episode Objectives:
    > 
    >    *   Explain what a library is and what libraries are used for.
    >    *
    >    *  Import a Python library and use the functions it contains.

* Lists appear as second-level lists (open circles). It shouldn't cause any troubles but I checked the PR  against `python-novice-inflammation` lesson only.
* (talking point) Output and Error blocks: Do we need these in Etherpads? Can you add lines "Output:" and "Error:" before them in case we decide they're useful?
* Some images still appear on the page (images within links). They don't appear on the Etherpad when copied but it would be good to detect-and-hide them too.
* Can you get rid of (all) non-essential formatting? You should be able to do that by stripping off  classes/ids/etc.

Denubis:
@maxim-belkin Argh, sorry, this slipped through my notifications. Can you link me the lesson you used as QA? Clearly the set of stuff exposed by instructor-training isn't the full set of stuff.

maxim-belkin:
I checked it by building the site locally but I can create a new repo to demonstrate this. I'll post a link to the repo in a bit

maxim-belkin:
https://maxim-belkin.github.io/swc-py-etherpad-test

fmichonneau:
Hi @Denubis 

the lesson infrastructure committee just met and we discussed this feature. At this time, it seems that even if it's not perfect, it would be useful for trainers teaching instructor training, and an improvement to the current workflow of manually checking the template against the teaching materials.

@maxim-belkin highlighted some potential areas of improvement. If you have the time to look into them and see if there are some you could implement now it'd be great. Otherwise, let us know and we'll merge as is until someone has time (and knowledge) to look into this more.

Denubis:
Hi @fmichonneau sorry, it's on my todo list to fix these improvements, I've just been swamped with teaching. Literally I look at these notifications (in my rss reader which I forward github stuff to) every day and feel guilty. Sorry! I'll ... see what I can do tonight to fix things.

Denubis:
> * Episode Questions and Episode Comments appear right underneath "Episode: Title", so there is no need to say "Episode" again. Moreover, I think that "Episode" can be removed from everywhere. It's not important how we call it (episode, lesson, theme, act, _etc._).

Done.

> * Lists (of objectives and questions) are spaced with empty items when copied to etherpad:
>   > Episode Objectives:
>   > 
>   > * Explain what a library is and what libraries are used for.
>   > 
>   > * Import a Python library and use the functions it contains.

Can't replicate on yours or mine. Checked on Chrome and Firefox. @fmichonneau can you replicate?
Copied loops in from python and it looks like it was conflating empty ordered lists in. I think I fixed it, but it's hard to test on instructor training alone.


> * Lists appear as second-level lists (open circles). It shouldn't cause any troubles but I checked the PR  against `python-novice-inflammation` lesson only.

Can't replicate locally, but that doesn't mean much. I'll spin up a browserstack instance if you have repro instructions.

> * (talking point) Output and Error blocks: Do we need these in Etherpads? Can you add lines "Output:" and "Error:" before them in case we decide they're useful?

Ah, that's a pythonic v. instructor training miss. I need to destroy code and output blocks. (Easy enough to readd them as we move this out from instructor training). Code added on line 110, happily all of the code blocks are .highlighter-rouge, so it's easy enough to replace.

> * Some images still appear on the page (images within links). They don't appear on the Etherpad when copied but it would be good to detect-and-hide them too.

For images, instead of killing them, I think I'm going to transform them into links, since I know I usually need to pop them into new tabs and that'll be a good reference. Right now it looks like images are absolute paths? This may be flaky if they're relative paths, but on the one test thing I looked at, they were absolute. This probably could use more testing.

> * Can you get rid of (all) non-essential formatting? You should be able to do that by stripping off  classes/ids/etc.

Removed all classes.

maxim-belkin:
> > * Lists (of objectives and questions) are spaced with empty items when copied to etherpad:
> >   > Episode Objectives:
> >   > 
> >   > * Explain what a library is and what libraries are used for.
> >   > * Import a Python library and use the functions it contains.
> 
> Can't replicate on yours or mine. Checked on Chrome and Firefox. @fmichonneau can you replicate?

The problem is still present. To replicate, go to https://maxim-belkin.github.io/swc-py-etherpad-test/, then Extras -> Etherpad, select "Episode objectives" and paste them to any Etherpad. I use https://pad.carpentries.org/test-issue1010 -- I left the output there on lines 56 -- 73:

image


> > * Lists appear as second-level lists (open circles). It shouldn't cause any troubles but I checked the PR  against `python-novice-inflammation` lesson only.
> 
> Can't replicate locally, but that doesn't mean much. I'll spin up a browserstack instance if you have repro instructions.

I'm not sure how exactly you're debugging this **locally** -- do you have a local Etherpad instance? I'm talking about the things I see when I copy them to Etherpad. 

See the screenshot above (or the Etherpad). Open circles == second-level lists.

image

Denubis:
Hi @maxim-belkin 

Here is my attempt to repro:
1) Chrome, ubuntu.
2) I go to https://maxim-belkin.github.io/swc-py-etherpad-test/etherpad/index.html
3) I hit ctrl-A, ctrl-C. (Debugging javascript other folk have written is entirely out of scope for this issue.)
4) I go to https://pad.carpentries.org/test-issue1010-bbs
5) I hit ctrl-V
6) Cannot reproduce.

Thus, I'm going to say that the error exists in your javascript. When I say locally, I mean using jekyll to serve the lesson locally. I recommend that if there exist options for javascript instructions, we carefully remove them until we can test further. However, being mindful of this, I'll figure out how to add a floating copy-to-clipboard button now.

maxim-belkin:
I use Firefox and was able to reproduce the problem in it on Windows and macOS.
Copying works correctly in Google Chrome (both Windows and macOS) and Internet Explorer (Windows). So, the problem is Firefox-specific. Going to try the button

maxim-belkin:
Same result with the copy button.

some (not all) code blocks, when copied, lose new lines so code appears to be on the same line.
Repro:
https://maxim-belkin.github.io/swc-py-etherpad-test/etherpad/index.html
Copy text that begins with "Let’s now change" and up to "Image: Updating a Variable"

Denubis:
Ok. Trying again. 

Going to browserstack, launching win 10 chrome 76.
Going to https://maxim-belkin.github.io/swc-py-etherpad-test/etherpad/index.html (again noting that this release is *not* for non-instructor-training lessons)
Click "copy to clipboard"
In browserstack, open new tab, Go to https://pad.carpentries.org/test-1010-bbs-3
Select all
Paste.
Looks fine. 
Starting firefox 68.
Click copy to clipboard.
Go to test-1010-bbs-4
select all

Ok. Finally reproduced. This seems like a firefox/windows bug. Trying to download/upload file.
Going to template.
Chose save file.
Going to 1010-bbs-4, 
Import has same issue.

Testing with instructor training: 
Browserstack, Windows 10 , Firefox 68. Going to https://denubis.github.io/instructor-training/, copy to clipboard, go back to test-1010-bbs-4. Reproduced. Testing with test-5 to make sure it's not a copy-over error. 

Ok, I've managed to reproduce the bullet issue on browserstack win 10 Firefox 68 on instructor training. Let me see what I can do.

Denubis:
Just for future reference of all of us. The bug was in calling the liquid filter | markdownify while inside an html list environment. (insert shrug emoji here.) Now trying to fix the code block problem.

Denubis:
Ok. Tested on firefox 68 on browserstack. Fixed. Tried to make code more readable though we may want to bring that into a new discussion for moving into the lesson template.

maxim-belkin:
The issue with stray `<li>` elements is gone, so  🎉   They do appear as level two ones, but this is not a big deal.

And I have an idea: since you've created that button - can't we just add a similar button to the navbar that would copy the current page (content of the `.container` div) to the clipboard?


Related to this PR: When stripping classes, could you please preserve the "container" class? It sets left/right margins for the body of the page.

Denubis:
Wasn't able to repro the level 2 thing, even in browserstack, win10, firefox 68. Going to give it a quick try in OS X. (see fresh test environment test-1010-bbs-osx).

Point two sounds like a great thing to raise in an issue after we start the process of integrating this into the lesson-template (which is still out of scope for the present commit)

Preserved .container as per request. Are you prepared to sign off on the inclusion of this in the instructor-training repo? At which point we can then start the conversation over on lesson-template to generalise this tool more widely?

fmichonneau:
Thanks all for your work on this! I'm going to merge as is, and we can open issues for additional functionalities and improvements.

PR #1013, opened 2019-07-15
suggestion  to fix #972

### LINKED ISSUE #972: Should OpenRefine (LC) be a suggested lesson for demos?
I just used it for a demo following the guidelines. The only recommended lesson is "Importing data into OpenRefine" but it's a super short lesson - even too short for the 5 minute teaching demo.

maneesha:
Adds an note in general, not just to this lesson, that trainees can continue to the next episode on if they have a very short episode.


ChristinaLK:
Good addition! I ran into this in the last demo I did...

PR #1015, opened 2019-07-17
Part 3 in checkout: add missing words

laurentheirendt:
This PR polishes the text in part 3 of the checkout.

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -224,9 +224,7 @@ and live code as if your computer was plugged into a projector
 and you were standing in front of a class.
 You may use whatever notes you want,
 provided they are not visible on your screen.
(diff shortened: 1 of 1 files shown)


karenword:
Thank you for the suggestion @laurentheirendt ! I agree that the language in this section could benefit from some clarification as "you are to" is uncommon in American usage. However, as a community of volunteers, we tend to steer clear of commanding words like "allowed" and prefer language that asks nicely instead. :) An alternative formulation more consistent with that philosophy might be "Please teach as you would in an actual workshop." 

If you would like to edit along these lines, you are welcome to do so!

laurentheirendt:
Thanks @karenword - it is done. 👍

raynamharris:
Thanks @laurentheirendt

PR #1016, opened 2019-07-17
Fix demopad link

sarah-peter:
In the step 3, the link to the Etherpad direct to a non-working pad (not loading). In the chat of that pad is the link to a new, properly working pad. I put the new link to the working etherpad.


ChristinaLK:
Thanks @sarah-peter !!

PR #1017, opened 2019-07-18
Fixing sub ordered list to proper markdown

Denubis:
Markdown is depressingly sensitive to spaces in its ordered lists. Converted the sub-list from an intended outer numbered list to an inner numbered list so it's 1. 1.i.


ChristinaLK:
Ah, thank you! I have definitely run into this problem before.

PR #1026, opened 2019-07-26
update link to new etherpad for teaching demos

raynamharris:
updates the link to the demo etherpad in the checkout lesson.

(Note, it was already updated in the more detailed checkout extras page here https://github.com/carpentries/instructor-training/pull/1016)


karenword:
We will be getting /demos back shortly. Will update all links when that happens!

raynamharris:
Ah, okay. Just to confirm, eventually, we will move to https://pad.carpentries.org/teaching-demos rather than https://pad.carpentries.org/teaching-demos-new (aka. without the -recovered or -new suffix).

karenword:
Yes -- this reversion is basically ready to go I just need to set aside the time to change All The Links. There is a chance we might be getting /trainers back today, so I might wait and do them all at once.

PR #1030, opened 2019-07-28
Fixed broken link to icebreaker questions.

mikblack:
Needs to be:
https://carpentries.github.io/instructor-training/icebreakers/index.html
instead of:
https://carpentries.github.io/icebreakers/index.html


karenword:
Thanks @mikblack !

Issue #1019, opened 2019-07-18
List of lessons to teach from


maneesha:
In the [Wrap up and Homework](https://carpentries.github.io/instructor-training/12-homework/index.html) section it lists a few SWC, LC, and DC lessons to choose from.  It says that:

> The episodes we recommend were listed in the welcome message sent to you before the start of this course,....

However the [welcome message](https://docs.carpentries.org/topic_folders/instructor_training/email_templates_admin.html#one-week-trainee-email-online-training) that trainees get says to review any of the lessons and does not make specific recommendations. Should we switch the curriculum material to not give specific recommendations or should we switch the welcome message to give specific recommendations?

ErinBecker:
Thanks for catching this @maneesha - all documentation should be updated to not make specific recommendations about which lessons to use.

raynamharris:
I think we want to leave some episodes there to help learners chose an episode (especially if they aren't very familiar with The Carpentries). Is it okay to change the wording (removing "recommended") but leave the specific examples?

maneesha:
I hadn't gotten to this since @ErinBecker  said we should not be making specific recommendations.  I also understand @raynamharris 's point about providing some guidance.   I like the idea of giving them some guidance but I'm not sure what the guidance would be based on.  If someone was most comfortable teaching a lesson other than whats [currently listed here](https://carpentries.github.io/instructor-training/12-homework/index.html) we may be doing them a disservice -- suggesting other lessons may pressure them into choosing something they are less comfortable with.

raynamharris:
There are 2 PRs open for this now (#1028 #1031 ). Both PRs say that learners can choose any episode, but the first PR also points the learners to the table of suggested demo lessons in case they need assistance. 

I'm thinking that the wording of both can be merged or there should be a group decision that no guidance should be provided.  Here's a snapshot of the proposed text. 

#1028 
Prepare for the live coding exercises. If you haven't already, pick an episode from an existing Software Carpentry, Data Carpentry, or Library Carpentry lesson and read through it carefully. Tomorrow, you will use this to practice live coding for 3 minutes in groups of three. Pick an episode from an actively maintained Software Carpentry, Data Carpentry, or Library Carpentry lesson. 
Read through the episode carefully and prepare to teach 3 minutes-worth of the material using live coding in small groups. Your group members will comment on the delivery and content.

[...]

PR #1031, opened 2019-07-29
trainees can choose any lesson - suggestion for #1019

maneesha:
This addresses #1019 by telling trainees they can choose from *any* lesson.


zpainter:
Honestly, I had my first experience with spreadsheets as a demo lesson last week. I didn't like it at all for a demo, and I think we should probably pare down the lessons we offer for both demos and the practice in instructor training.

Denubis:
Linking this to #1156 since this does seem to be linked to recent convos.

PR #1036, opened 2019-08-08
[translations] Clean lessons to create po

dpshelio:
The `po` files are the tokenised version of the lessons. The tool we are using complains if the following *typos* or empty lines at the end of sections are not fixed.


dpshelio:
Thanks @ChristinaLK

PR #1038, opened 2019-08-14
Add plyr lesson to list "avoid"

jcoliver:
The plyr package is largely deprecated, with dplyr and purrr being the replacements. There are plans to retire the plyr episode soon, so we should not expect instructors in training to instruct this episode.


ChristinaLK:
thanks @jcoliver !

raynamharris:
Thanks @jcoliver for this issue for contributing to the discussion about replacing this episode altogether in https://github.com/swcarpentry/r-novice-gapminder/issues/540

PR #1042, opened 2019-08-29
Add person for timekeeping

konrad:
This make it explicit that somebody should be responsible for time keeping during the teaching exercise. 

This needs also equilateral modifications in the etherpad template.


karenword:
great addition! Thanks @konrad ! I've also updated the etherpad with your change.

PR #1045, opened 2019-09-22
Missing word fix

anenadic:
Fixed a missing word


karenword:
Thanks @anenadic -- @raynamharris and I have reviewed this and think the language is actually ok as-is. "Live coding" is generally used as an action throughout our curriculum, and existing language is consistent with that usage.

PR #1046, opened 2019-09-24
Added links to Handbook in Key Points

anenadic:
Added links to Handbook in Key Points too so people can refer to them more easily.


karenword:
Thanks @anenadic !

PR #1047, opened 2019-09-24
Included Library Carpentry when talking about lesson format

anenadic:
Included Library Carpentry when talking about Software and Data Carpentry lesson format. An alternative would be to use "The Carpentries' lesson format", to include all current and future lesson organisations. and avoid the need to updating this.


raynamharris:
Thanks @anenadic!

PR #1049, opened 2019-09-30
Fix link to icebreakers

gcapes:
This didn't give the correct link for a local build,
even though it gets magically fixed for github pages.

It gave an extra 'instructor-training' in the path.


karenword:
Thanks @gcapes !

PR #1050, opened 2019-10-03
Added a link to example good first contributions in checkout process.

erjank:


brownsarahm:
This should be linked also on the [extras/checkout](https://github.com/carpentries/instructor-training/blob/gh-pages/_extras/checkout.md) page. 

Possibly also replacing the links to the lesson program pages and providing one per lesson instead (could be more maintenance work, but...)

erjank:
Whoops, also wrapping the http needs to be () not []

karenword:
Thanks @erjank !

PR #1052, opened 2019-10-08
Use more realistic praise example

gcapes:
This sounded very unnatural (at least to a UK audience).
It also contrasts nicely with the previous example, and shows the subtle difference in language promoting different mindsets.

### DIFF
--- a/_episodes/09-mindset.md
+++ b/_episodes/09-mindset.md
@@ -97,12 +97,14 @@ highlighting our ability to improve with effort.
 > - You're getting really good at that. Keep up the hard work!
 > - You're really good at that.
 > - That was a hard problem. You didn't get the right answer, but look at how much you learned trying to solve it!
+> - You're a natural!
 >
 >> ## Solution
 >> - Effort-based.
 >> - Improvement-based.
 >> - Performance-based.
 >> - Improvement-based.
+>> - Performance-based. Explicit reference to 'intrinsic' ability.
 > {: .solution}
 {: .challenge}
 
--- a/_episodes/09-mindset.md
+++ b/_episodes/09-mindset.md
@@ -95,7 +95,7 @@ highlighting our ability to improve with effort.
 >
 > - I like the way you tried a couple of different strategies to solve that problem.
 > - You're getting really good at that. Keep up the hard work!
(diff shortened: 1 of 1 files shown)


raynamharris:
Thanks @gcapes. You bring up a very good point. How slightly modifying your suggestion to say "You're naturally good at that."

gcapes:
I'm afraid that also sounds rather unnatural to me. 

This is an example of 'performance based' praise. I don't think there's a need to imply that the origin of the ability intrinsic, is there?

karenword:
I think the question here is how subtle we want this prompt to be -- maybe we could consider *adding* this as a prompt rather than replacing the reference to intrinsic talent?

I can see the value in encouraging people to examine the difference between saying "You're really good at that" and "You're *getting* really good at that" but that also relies on some deeper thought about the different ways that people might interpret the former. It's not immediately apparent that this kind of performance-based praise could support the belief that their performance is based on ability rather than effort (fixed mindset), but this is the negative outcome that research suggests we could avoid. 

In making the connection between praise and mindset, the most straightforward examples will include intrinsic references ("talented" or "natural"). A more challenging example will ask learners to consider how praise that doesn't directly reference intrinsic talent might be interpreted that way by someone approaching the task with a fixed mindset.

gcapes:
Sure, we could add rather than replace *if* we can think of a natural-sounding example for intrinsic ability. The point of the exercise is to make people think about differences in wording, and how that can encourage the shift to growth mindset.

An unnatural sounding example doesn't elicit much thinking other than "I'd never say something like that anyway". Perhaps to a US audience the example sounds fine, but to a UK (/international?) audience it distracts from the point of the exercise.

tobyhodges:
I think the phrase "You're a natural!" would be familiar to a UK audience. Not sure how universal it would be, though.

elletjies:
> I think the phrase "You're a natural!" would be familiar to a UK audience. Not sure how universal it would be, though.

I think this will also resonate with a South Africa audience.

gcapes:
Thanks @tobyhodges  and @elletjies 
I'll update the PR.

karenword:
Thanks @gcapes !

PR #1053, opened 2019-10-08
Fix description in post-training survey link

mattipt:


karenword:
Thanks for the catch @mattipt !

Issue #1044, opened 2019-09-11
Key point needs to be updated.


amyehodge:
The first key point at the end of this episode (https://carpentries.github.io/instructor-training/21-carpentries/index.html) is "The Carpentry materials are all openly licensed, but Software and Data Carpentry names and logos are trademarked." This needs to be updated to include Library Carpentry.

raynamharris:
This has already been mentioned in PR https://github.com/carpentries/instructor-training/pull/1054. The question at hand, I think, is whether the list of trademarked logos should include all the lesson programs AND the Carpentries.

karenword:
This appears to have been addressed. Closing! :)

PR #1054, opened 2019-10-08
Added Library Carpentry to the list

anenadic:
Added Library Carpentry to the list alongside Software and Data Carpentry.

### DIFF
--- a/_episodes/21-carpentries.md
+++ b/_episodes/21-carpentries.md
@@ -11,7 +11,7 @@ objectives:
 - "Get connected with the Carpentry community."
 - "Describe where you can go to get information on running a workshop."
 keypoints:
-- "Carpentry materials are all openly licensed, but Software and Data Carpentry names and logos are trademarked."
+- "Carpentry materials are all openly licensed, but Software, Data and Library Carpentry names and logos are trademarked."
 - "Carpentry workshops must cover core concepts, have at least one certified instructor, and use our pre- and post-workshop surveys."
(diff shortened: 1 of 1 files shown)


karenword:
Thanks @anenadic ! I wonder if we should include The Carpentries here now, too, and possibly adjust the language a bit more to accommodate that. @maneesha I wonder if you can weigh in on what would be the most fully accurate and least clunky way to describe all this as it applies to our collection of trademarked things?

raynamharris:
This issue has come up again in https://github.com/carpentries/instructor-training/issues/1044. Should LC and the Carpentries be added?

maneesha:
I am not 100% sure if all the logos are trademarked. I think 
@Talishask  can offer input from the Business team perspective.

Talishask:
Hi @maneesha, 
Thanks for tagging me here and apologies on the delay. Not all of our logos are trademarked. I can confirm that both DC & SWC have trademarks but not LC or The Carpentries in general.

Due to transitions and then COVID-19 we postponed the trademarking project for the business team.

Let me know if you need anything else

brownsarahm:
Should we update the keypoint to say that those two are explicitly not yet trademarked? or leave it as is for now?

ChristinaLK:
I think closing this PR and leaving as is might be good? because the existing text is accurate.

karenword:
Thanks @Talishask -- whenever trademarks do happen, would you please let us know so we can update? I will go ahead and close this PR for now.

Talishask:
Yes! certainly. I've created an Asana task within the trademark project to remind me to update this page once we've updated/added trademarks.

PR #1064, opened 2019-11-13
Expert awareness gap

brownsarahm:
remove references to seeing in blind spot introduction, make it focus on change in thought. 

add a box with thoughts on the term expert awareness gap


karenword:
@brownsarahm thank you! This looks like a really graceful solution to the problem raised in the Trainers Slack. It acknowledges the existing terminology and tells a story of our community's efforts to update ableist language in a thoughtful way. 

I think we should leave this open for comment for a bit, since it adjusts how we present one of our core concepts, which is not something we would normally do outside of a major overhaul.

dvanic:
@karenword @brownsarahm I think it's a great way to handle a challenging topic!!! I absolutely love this piece.

The only caveat I have is this phrase:

> thus avoiding abelist language

I did a quick survey around the office (mostly native Australian English or British English speakers, from a broad diversity of ethnic/cultural backgrounds), and most of them were not familiar with the term "ableist". This specific term is not used nearly as often as it is in the US in discussions around disability (so the concept is discussed but without a specific encapsulating word).

To keep our curriculum accessible internationally, especially for people for whom English is a second language, could this phrase paraphrased somehow, keeping the meaning accessible but not using the term?

brownsarahm:
@dvanic good catch. I tried a rephrase, let me know it that's better.

mikemhenry:
I think we should also update `_episodes/24-practices.md` (line 45), `_extras/glossary.md` (line 66), and `_episodes/08-motivation.md` (line 99) as well as part of this PR.

brownsarahm:
thanks @mikemhenry ! done and done!

brownsarahm:
@kariljordan done!

Issue #981, opened 2019-05-30
more concept map examples


brownsarahm:
Tabulating a list of them here, to be placed somewhere more permanent/decide how to use them later

https://twitter.com/dcsoto_cl/status/1111447971550228480

https://twitter.com/noccaea/status/1121116063977082881

https://twitter.com/StatalieT/status/1128209011814424576

Not directly, but an application: https://twitter.com/CIRCA_StAndrews/status/979294324436463616

brownsarahm:
https://twitter.com/Tiana_Athriel/status/1123443249681711104

https://twitter.com/morganoneka/status/1136045253033713664

ChristinaLK:
Great, thanks @brownsarahm !

brownsarahm:
https://twitter.com/BeaCurious/status/1151490842617475072?s=19

orchid00:
These are all great examples! thanks @brownsarahm for collecting them and sharing.

I'd like to point out.

- Cognitive maps are the most general type of mental-model visualization. No restrictions on structure or form. These can be dived in two types:
    - A mind map is a tree that represents a central topic and its subtopics. There's no distinction between different types of relationships among nodes.
    - Concept maps have a defined structure where have emphasis on relationships between concepts (nodes), and a node can have multiple parents. Usually read top down. Usually a concept is a word and the relationships are action verbs. 


[This article ](https://www.nngroup.com/articles/cognitive-mind-concept/) helped me understand the difference. It also came with a [very short video](https://www.nngroup.com/videos/visual-mapping-strategies/) I'd recommend watching.

raynamharris:
what do you think about this @brownsarahm https://github.com/carpentries/instructor-training/pull/1067

![Screen Shot 2019-11-17 at 2 31 18 PM](https://user-images.githubusercontent.com/7544197/69015246-0df96e80-0947-11ea-8644-f44759cf5dd0.png)

brownsarahm:
this is great!

PR #1067, opened 2019-11-17
Create fromthelearners.md

raynamharris:
As per the discussion in https://github.com/carpentries/instructor-training/issues/981, this is a place to collect graphical summaries from learners in our workshops. I think its more appropriate to have it as an "extra" page rather than including these somewhere in the official curriculum.


raynamharris:
The tweets are rendered like this

Screen Shot 2019-11-17 at 2 31 18 PM

PR #1073, opened 2019-12-16
Fix AMY links to point to amy.carpentries.org.

mattipt:
AMY seems to have moved from https://amy.software-carpentry.org to https://amy.carpentries.org. Fix links accordingly.


fmichonneau:
Thanks!

PR #1074, opened 2019-12-21
Capitalizing "carpentries" in venn diagram

rachellombardi:
I believe that the word "carpentries" located in the white space of the venn diagram should be capitalized to be "Carpentries".


brownsarahm:
Thanks @rachellombardi for pointing this out.  I've created an issue for this instead of a pull request.  Accepting (merging) this pull request would change text in the lesson instead of getting someone with the ability to change the diagram.

PR #1077, opened 2020-01-16
more updates of expert blind spot to awareness gap

brownsarahm:
objectives and keypoints were not updated yet


karenword:
Thank you @brownsarahm !

Issue #874, opened 2019-01-14
Key points for "Wrap-Up and Homework for Tomorrow" lesson


anenadic:
"Wrap-Up and Homework for Tomorrow" lesson has no key points (which was probably the intention) but they appear at the bottom as an empty green box with a single empty bullet point. So we should either add key points (if any) or look into checking that empty key points box does not appear if possible (if not - it is OK).

anenadic:
I think the file to have a look at is in _includes/all_keypoints.md and the code should look something like (this is not correct syntax probably):
{% if episode.keypoints not empty %}      
       <td class="col-md-9">
        <ul>
        {% for keypoint in episode.keypoints %}
        <li>{{ keypoint|markdownify }}</li>
        {% endfor %}
        </ul>
      </td>
{% endif %}

ChristinaLK:
is this change worth making in the upstream lesson template, not just locally?

@rgaiacs @fmichonneau

PR #1081, opened 2020-02-18
Adding regex as a pure avoid for LC

Denubis:
Just had to manually check through the data intro lesson, so I've added it as a "pure avoid."


zpainter:
Perhaps the spreadsheets (tidy data) wouldn't be truly terrible compared to anything else on the docket, although I would agree that this lesson family in general should be avoided for demos. It is very domain and audience specific.

_extras/demo_lessons.md
```diff
@@ -65,7 +65,7 @@ If you are an instructor in training and wish to use lessons from Data Carpentry
...
 | [Library Carpentry: OpenRefine](https://librarycarpentry.org/lc-open-refine/) | [Importing data into OpenRefine](https://librarycarpentry.org/lc-open-refine/02-importing-data/index.html) | any other episode | [Introduction to OpenRefine](https://librarycarpentry.org/lc-open-refine/01-introduction/index.html) has no live coding. <br/> Later episodes have dependencies. |
-
+| [Library Carpentry: Library Carpentry: Introduction to Working with Data (Regular Expressions)](https://librarycarpentry.org/lc-data-intro/) | No episodes | Avoid all episodes. | Episodes are entirely exercises and a poor fit for live coding demos |
```
brownsarahm, line 68:
```suggestion
| Library Carpentry: Introduction to Working with Data (Regular Expressions)](https://librarycarpentry.org/lc-data-intro/) | No episodes | Avoid all episodes. | Episodes are reference with exercises and a poor fit for live coding demos |
```

brownsarahm:
@zpainter I think adding the tidy data as an allowable lesson is likely a good idea, but separate from avoiding this lesson.  That is related to #874 which has slowed discussion, but maybe is worth reviving over there and letting this get merged in, since it is a core lesson that we want trainees to avoid.

_extras/demo_lessons.md
Denubis: I tried another formulation of the avoid. Fixed the double header. I'm going to resolve this convo. Take a look at my edit to see if it works?

brownsarahm:
LGTM, thanks! @Denubis

PR #1083, opened 2020-03-01
Fixing broken link

jduckles:
There is no local variable to the page `training_site` it is set in `_config.yml` and scoped to `site.training_site`. Should fix the link.


karenword:
Thanks @jduckles !

PR #1084, opened 2020-03-02
replaced phrase on errors in ep 8 line 36

ragamouf:
I suggest to replace "when done well" with "when handled transparently" in Line 36. "Done well" is a good example of performance-based feedback, which was cautioned in previous instructor episodes because it is generic in nature. 

I understand the point of this section in instructor training is to normalise and make space for errors in the workshop, which is done by handling errors transparently and honestly. I realise this is quite a minor tweak, but I believe it would help the overall flow of the episode.


karenword:
Excellent point -- thanks @ragamouf !

PR #1085, opened 2020-03-03
Reworded "blind spot" to "expert awareness gap"

stewartmacdonald:
I don't know what I'm doing in GitHub, sorry. Just trying to tidy up some of the wording in this lesson. This lesson explicitly says "we prefer not to use the phrase 'blind spot'", but  then goes on to use that phrase. I've suggested some alternative wording.


karenword:
Thank you @stewartmacdonald ! And welcome - nicely done! :)

PR #1091, opened 2020-03-10
remove attendance policy

sheraaronhurt:
we do not collect attendance

### DIFF
--- a/_episodes/21-carpentries.md
+++ b/_episodes/21-carpentries.md
@@ -105,7 +105,7 @@ You may only call a workshop a Software Carpentry, Data Carpentry, or Library Ca
 
 *   it covers the core topics (see below),
 *   at least one of the instructors is a certified Carpentries Instructor,
-*   you run our standardized pre- and post-workshop assessments and ensure everyone participates, and
+*   you run our standardized pre- and post-workshop assessments and ensure everyone participates.
 
 
 ### Materials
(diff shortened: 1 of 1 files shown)


Denubis:
It may be worth noting that many local hosts *do* track attendance, either for compliance or local stats or institutional goals -- and that this is a specific matter to liaise with the host about? (And it's really fun when the physical sign-in sheets get lost...)

brownsarahm:
Helping the host collect attendance info is listed in the [instructor checklist in the handbook](https://docs.carpentries.org/topic_folders/hosts_instructors/hosts_instructors_checklist.html#instructor-checklist) and that is a part of this episode later on. Is that enough emphasis on that @Denubis?

_episodes/21-carpentries.md
```diff
@@ -106,7 +106,7 @@ You may only call a workshop a Software Carpentry, Data Carpentry, or Library Ca
 *   it covers the core topics (see below),
 *   at least one of the instructors is a certified Carpentries Instructor,
 *   you run our standardized pre- and post-workshop assessments and ensure everyone participates, and
```
alex-konovalov, line 108: This sentence is now incomplete.

brownsarahm:
```suggestion
*   you run our standardized pre- and post-workshop assessments and ensure everyone participates.
```

sheraaronhurt:
@brownsarahm thanks for noting that in the instructor checklist. I will get that removed from the handbook.

karenword:
I think in this particular case, since this is about branding, we should go ahead and remove the language regarding attendance. @sheraaronhurt I think you might consider leaving the checklist in the handbook as-is, since that specifically refers to supporting the host and does not suggest that attendance records are required by us.

PR #1098, opened 2020-03-20
add pronouns to sign in, make email optional

fmichonneau:
make sure we ask participants their pronouns when signing in in the Etherpad, and make clean we don't require them to share their email address with everyone publicly.


karenword:
Thanks @fmichonneau !

PR #1107, opened 2020-04-14
[fix #1106] upgrade to latest script to include figures

### LINKED ISSUE #1106: Figures missing from the Figures page
I don't know enough about [the script](https://github.com/carpentries/instructor-training/blob/gh-pages/_extras/figures.md) that generates the [Figures page](https://carpentries.github.io/instructor-training/figures/index.html) to understand what's going on, but it looks like it's not picking up all the figures. 

Having all the figures on this page would be extremely helpful for running through my first Instructor Training!

fmichonneau:
Brings the latest version of the script to render all images in a page.

### DIFF
--- a/_extras/figures.md
+++ b/_extras/figures.md
@@ -1,40 +1,79 @@
 ---
-layout: page
 title: Figures
 ---
+
+{% include base_path.html %}
+{% include manual_episode_order.html %}
+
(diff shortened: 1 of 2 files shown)


karenword:
Thank you @fmichonneau !

PR #1116, opened 2020-04-26
fix duplicated word

profgiuseppe:
Removed duplicated word.


ChristinaLK:
thanks @profgiuseppe !

PR #1120, opened 2020-04-30
make sharing pronouns more inclusive

fmichonneau:
* using "preferred" isn't recommended
* add wording to make sure we don't force people to share their pronouns if they don't want to do it


ChristinaLK:
thanks @fmichonneau ! this looks good to me. @karenword anything to add?

ChristinaLK:
eventually, I'd like to word this more as directed to the participants than the trainer but that can be a different PR. :)

PR #1123, opened 2020-05-13
Switch order of recommended contribution types

ErinBecker:
The [checkout page in extras](https://carpentries.github.io/instructor-training/checkout/index.html#lesson-change) lists possibly types of contributions, starting with the type that we are most encouraging from trainees. This modifies the same list in the lesson episode proper to match the order in the extras file.


ErinBecker:
@karenword - please review :-)

karenword:
Thanks @ErinBecker !

PR #1131, opened 2020-06-01
suggested questions for Trainers to prepare to co-teach

maneesha:
I have used this list of questions for two training events and it has been very useful for us as Trainers to prepare to teach, and be sure our expectations of each other are clear.


brownsarahm:
Thanks @maneesha !

Issue #1133, opened 2020-06-08
Teaching demo rubric and 95% pass stat


arieldeardorff:
I was recently providing feedback to a learner who did not pass their teaching demo and thought that the [demo rubric](https://carpentries.github.io/instructor-training/demos_rubric/) would be a good thing to share. However, I decided not to because the line in the second paragraph about how 95% of people pass. Could we change this to "the majority of people pass" or something like that? I am worried that this will make people feel bad if they are part of the only 5% that don't pass.

It would also be great to add "copying and pasting code from notes" (or something similar) under negative delivery. I think we talk around this but there isn't anything that explicitly says this would be a negative in the demo.

brownsarahm:
I think that these are great changes, can you submit a PR?

arieldeardorff:
Pull request #1134 submitted!

PR #1134, opened 2020-06-08
Updates to teaching demo rubric

arieldeardorff:
As discussed in issue #1133 I updated the language around passing demos and added another line to the rubric about typing out all of their code.


brownsarahm:
Looks great to me!

Issue #1108, opened 2020-04-16
Connect CoC information to guidelines in handbook


maneesha:
In [Managing a Diverse Classroom](https://carpentries.github.io/instructor-training/18-management/index.html), the CoC section should link to the guidelines in the handbook.  The recommendations in the IT curriculum should also align with the recommendations in the handbook. For example the curriculum says: 

> If you are an instructor, and believe that someone in a workshop has violated the Code of Conduct, you may warn them, ask them to apologize, and/or ask them to leave the workshop, depending on the severity of the incident and your comfort level.

However, the [Code of Conduct response guidelines](https://docs.carpentries.org/topic_folders/policies/incident-response.html#checklists-for-responding-to-an-incident)  recommend *not* asking the person to apologize:

> Do not allow the reported person to make an apology to the reporter or impacted person. Often an apology centers the reported person’s feelings and not the person who was impacted. You may accept their apology and offer to pass it on, but you’re not required to if you think it would negatively impact the reporter.

karinlag:
I just want to point out that these two situations are slightly different. The response guidelines are more geared towards someone taking a CoC report. That is, a situation where one person is one-on-one talking with a person who wants to report that something has happened. I think, at least as I read the Diverse Classroom section, that that more deals with a situation where something has happened in full public view during a workshop. If so, I think that asking for someone to publicly apologize might be in order. However, the training materials and the CoC materials should become more aligned in this area.

PR #1138, opened 2020-06-18
Remove "ask to apologize"

karenword:
Addresses https://github.com/carpentries/instructor-training/issues/1108 -- while this context may differ from that in our Handbook, I think it is probably best to remove the suggestion that Instructors demand an apology. This is not an especially useful strategy (at best) and the appearance of inconsistency is a problem.


ChristinaLK:
agreed - this looks good.

PR #1145, opened 2020-06-19
Fixing exporter because relative_root_path was silently removed

Denubis:
@brownsarahm pinged me on slack notifying me that all my links were broken. 

I verified with the following reproduction:
```
I go to: https://carpentries.github.io/instructor-training/etherpad/index.html
click copy to clipboard. Go to https://pad.carpentries.org/2020-06-19-bbs-bug-exporter-test-1, and hit paste. Bug reproduced.
```
Cause: "Relative_root_path" does not seem to be a supported variable any more, but was dropped without catching it throughout the repo?

Fix. use the liquid "| absolute_url" filter to not need to worry about baseurl or site.url stuff. Looking at https://denubis.github.io/instructor-training/etherpad/index.html it looks like all lesson urls are fixed. Image urls are not manipulated by jekyll. I just tested one and it still works.


Denubis:
@fmichonneau It might be worth grepping for relative_root_path to see what other relics have been left behind?

fmichonneau:
I can't see why the `relative_root_path` variable would have been removed... (it looks like it's still here: https://github.com/carpentries/instructor-training/blob/gh-pages/_includes/base_path.html).

I think the issue you have here has a different origin. There are some cases where the order of evaluation makes this variable not work the way it's supposed to. We have found in the past that using ``{{ page.root  }}/{{ relative_root_path }}` is actually the correct incantation. `absolute_url` doesn't work when trying to make the website offline-capable but if it works for the Etherpad exporter, I'm happy to merge this PR

Denubis:
@fmichonneau is there a local jekyll config file that spins up these variables? The term "order of evaluation" is a terrifying statement. Out of curiosity, what do you mean by "offline" mode? Local Jekyll?

fmichonneau:
no, this particular variable gets created in the context of the `include` statement.
here is the relevant Jekyll section about the order of interpretation https://jekyllrb.com/tutorials/orderofinterpretation/#order-of-interpretations
for offline mode, I mean the output of `bundle exec jekyll site`. We're not 100% there but ideally we'd want the output in the `_site` to be put on a USB drive (without having a webserver running in the background) and have this static site to work.

brownsarahm:
@Denubis @fmichonneau is this good to be merged?

Denubis:
@fmichonneau do we want to hold off on offline forking for later? Or do you want me to try to figure out how to get the above stuff working in _extras?

Denubis:
@fmichonneau 
Running it with jeykll serve `<h1>{{episode.title}} <br/><a href="{{ page.root }}/{{ relative_root_path }}{{episode.url}}">{{ page.root }}/{{ relative_root_path }}{{episode.ur}l}</a></h1>`

 produces 
```Welcome
..//01-welcome/index.html``` -- so I think this is going to be problematic for offline patterns... 

Testing on my local github pages... 
https://denubis.github.io/instructor-training/etherpad/index.html produces the same. 

I think you and I need to chat on slack and figure out what interesting things are happening here. Reverting to my prior instance. 

@brownsarahm As soon as I revert it will be good to commit.

brownsarahm:
Thanks @Denubis !

Issue #1143, opened 2020-06-18
Direct policy information to Handbook


maneesha:
Currently we have this sentence 

> Policies related to instructor training and workshops The Carpentries can be found in the policies repository. Please be sure to read through the instructor no-show policy before signing up for your first workshop.

which includes links to this policy repo:  https://github.com/carpentries/policies
This repo may be out of date.

Instead this should link to policy information in our handbook: https://docs.carpentries.org/topic_folders/policies/index.html

PR #1146, opened 2020-06-24
fix policies links

karenword:
Addresses https://github.com/carpentries/instructor-training/issues/1143


ChristinaLK:
review please @carpentries/instructor-training-maintainers

PR #1152, opened 2020-07-08
Excluding manipulating data because it has dependencies

Denubis:
`surveys <- read_csv("data_raw/portal_data_joined.csv")` depends on the prior lesson


ChristinaLK:
I want to make sure I understand correctly: the dependency in question is downloading the data file from figshare (which happens in the previous lesson)?

Denubis:
Yes. It assumes scripts/directory structure from the prior episode. Taking a look at https://datacarpentry.org/R-ecology-lesson/03-dplyr.html It has a huge amount of setup, and then abrutply loads data from https://datacarpentry.org/R-ecology-lesson/02-starting-with-data.html. Looking at this lesson more closely, it doesn't feel like the lesson in general is a good fit for demonstrating live coding, since there's a great deal of explanation/cognitive-setup at the start of episodes. (And it leaves us exactly one episode which is allowed).

ChristinaLK:
Got it - I think I've always assumed that people will have that directory structure/file if they prepped the lesson in advance, but I think you're right that (for now) it's a dependency to be avoided. I also agree w/ your point about the lengthy intro.

PR #1154, opened 2020-07-22
updated the article link

peranti:
The article was earlier redirected to the preprint version and the changes have been made to point towards the published version.


brownsarahm:
Thanks @peranti this looks good to me and since the published version is open access, I agree that version is better to link.

Issue #1124, opened 2020-05-17
Capitalize Carpentries in venn diagram


brownsarahm:
Per #1074 

In `fig/carpentries-venn-diagram_20190522.svg` the word "carpentries" located in the center white space of the diagram should be capitalized to be "Carpentries".

mikemhenry:
I can take care of this issue

mikemhenry:
This issue can be closed now :)

brownsarahm:
Closed with #1167

brownsarahm:
Thanks @mikemhenry ... i don't know why the ink didn't work to close it automatically

PR #1167, opened 2020-09-04
Fixes issue #1124

mikemhenry:
Capitalized the word carpentries to Carpentries located in the center
white space of the diagram. For future font hackers, the font name is
Adobe Courier and the font size is 8pt and it is bold :)


brownsarahm:
This looks good to me, thanks @mikemhenry !

@carpentries/instructor-training-maintainers I think noting this font research mike did somewhere more findable would be really good to future editors, but I'm not sure where that is...

ChristinaLK:
Do we want to create a lesson style guide, even if mainly points to the Carpentries one + then has lesson specific notes like this?

mikemhenry:
I love that idea!

PR #1170, opened 2020-09-23
add descriptive alt text to motivation episode

ErinBecker:
Adds descriptive alt text for the two images in the motivation and demotivation episode.


brownsarahm:
thanks erin!

ErinBecker:
Wow, that was fast! Thanks @brownsarahm!

PR #1182, opened 2020-10-20
expanding word contractions across instructor training content 

serahrono:
In reading up about web accessibility and content design for broad audiences, I learnt that negative word contractions might make readability and comprehension more difficult for some, and it is best to avoid these. See https://medium.com/@joanne.schofield/using-contractions-could-be-making-your-writing-inaccessible-4034bb3cd76a, for example. With this in mind, I have started to expand word contractions across content in this repository.


ChristinaLK:
Thanks @serahrono !

maneesha:
Thanks for taking this time to do this and enforcing these guidelines in our materials @serahrono 
If only the sources you quoted would avoid using words like *I've, couldn't, doesn't, don't*, in the same documents where they say not to do so.  :roll_eyes:

serahrono:
The jury is still out on whether positive contractions like _you'll, it's, let's_ are ok to leave in text or not - for example, 
- https://readabilityguidelines.co.uk/grammar-points/contractions/ suggests positive contractions _may_ be fine, while
- https://gds.blog.gov.uk/2016/02/23/writing-content-for-everyone/ reiterates that punctuations in contractions can be hard to read and therefore slow people down.

My suggestion is for us to expand all contractions across our content. Thoughts?

ChristinaLK:
I'm in favor of expanding all contractions - it certainly doesn't HARM anything and it sounds like it would be helpful.

maneesha:
I have no strong feelings about making the change either way. My frustration was not with the changes proposed here, but rather with the sources we are citing -- they fail to follow their own recommendations (for positive or negative contractions).

brownsarahm:
Also in favor of all contractions expanded.

CONTRIBUTING.md
```diff
@@ -70,7 +70,7 @@ There are many ways to contribute,
 from writing new exercises and improving existing ones
 to updating or filling in the documentation
 and submitting [bug reports][issues]
-about things that don't work, aren't clear, or are missing.
+about things that do not work, are not clear, or are missing.
```
ChristinaLK, line 73: these changes should be propagated up to `styles`. I'll open an issue there.

ChristinaLK: Or not...where does the stock CONTRIBUTING.md file live?

_episodes/01-welcome.md
```diff
@@ -173,20 +173,20 @@ To orient yourself, there is a schedule on the workshop webpage which is linked
...
-of lesson design called "Reverse Instructional Design", but it won't be an emphasis
+of lesson design called "Reverse Instructional Design", but it will not be an emphasis
 of this particular course.
 
-If there's a particular topic that you would like us to address, let the trainers
+If there is a particular topic that you would like us to address, let the trainers
 know.
 
-## Let's Get Started
+## Let us Get Started
```
ChristinaLK, line 186: This sounds really funny uncontracted!

_episodes/02-practice-learning.md
```diff
@@ -299,8 +299,8 @@ but each of the other answers provides valuable insight.
 >> ## Solution
 >>
 >> *   If the child answers 32, they are throwing away the carry completely.
->> *   If they answer 312, they know that they can't just discard the carried '1',
->>     but doesn't understand that it's actually a ten
+>> *   If they answer 312, they know that they cannot just discard the carried '1',
+>>     but does not understand that it is actually a ten
```
ChristinaLK, line 303:
```suggestion
>>     but do not understand that it is actually a ten
```

_episodes/05-memory.md
```diff
@@ -243,7 +243,7 @@ Cognitive load theory's proponents claim that eliminating extraneous cognitive l
 way to do this is by using *guided practice*: we set up a structure in which learners can
 apply their skills with direction and get feedback on their progress.
 
-> ## Why We Don't Use Minimal Guidance
+> ## Why We Do not Use Minimal Guidance
```
ChristinaLK, line 246: should "not" be capitalized?

_episodes/20-checkout.md
```diff
@@ -94,12 +94,12 @@ be discussed before changes are merged.
...
 * **You get to vote!** Badged Instructors are eligible to vote in Carpentries Executive Council elections for their first year, and for all subsequent years in which they continue to participate through teaching or other involvement. See [this section from the Carpentries Bylaws](https://docs.carpentries.org/topic_folders/governance/bylaws.html#eligibility-rights-and-termination-for-voting-members) for rules governing continuing voting eligibility.
-* **You've got something to brag about!** Looking for language to add to showcase your skills and experience? [Here is some text](https://github.com/carpentries/commons/blob/master/text-for-instructors.md) to get you started!
+* **You have got something to brag about!** Looking for language to add to showcase your skills and experience? [Here is some text](https://github.com/carpentries/commons/blob/master/text-for-instructors.md) to get you started!
```
ChristinaLK, line 102:
```suggestion
* **You have something to brag about!** Looking for language to add to showcase your skills and experience? [Here is some text](https://github.com/carpentries/commons/blob/master/text-for-instructors.md) to get you started!
```

ChristinaLK (commented):
Some of these files originate in template repos and should probably be changed there as well.

CONTRIBUTING.md
tobyhodges: it is in `styles`, but in the `boilerplate` directory used to create these files when the lesson is initialised: https://github.com/carpentries/styles/blob/gh-pages/bin/boilerplate/CONTRIBUTING.md

_episodes/05-memory.md
brownsarahm: I was curious so I did a quick search... it seems like there isn't one decision on this.  If we want to be consistent somehow, we could pick and put it in the main style guide for the Carpentries on which flavour of title case we prefer.  My vote is that "do" and "not" are in the same case. 

https://www.thoughtco.com/which-words-in-a-title-should-be-capitalized-1691026

(PS. i'm very proud of myself for not using the US spelling of flavour, even though my browser is yelling at me for spelling it "incorrectly")

ChristinaLK: Agree that do + not are either both upper or lower case.

_episodes/01-welcome.md
karenword, line 186:
```suggestion
## Getting Started
```

karenword:
```suggestion
## Getting Started
```

brownsarahm:
Should we get this one merged in (and on data-lessons)? So that these changes are in before the big revisions?

ChristinaLK:
If we can, I think that would be good. @karenword  ?

PR #1185, opened 2020-10-21
fix urls by removing static prefix

murraycadzow:
I think this is the last page that still had broken links due to the carpentries url changing to no longer have "static"


brownsarahm:
Thanks!

PR #1187, opened 2020-10-23
Update 08-motivation.md

klbarnes20:
During Hacktoberfest, The Carpentries is removing contractions from their website for accessibility. Following their lead, I have removed contractions to make this page more accessible.


klbarnes20:
I've decide to close this as Serah has a much more thorough PR on this here: https://github.com/carpentries/instructor-training/pull/1182

PR #1188, opened 2020-10-26
add title element to testimonial

zkamvar:
The testimonial element only shows up as a pink box without the header. With the header, it's more clearly denoted as a testimonial.


brownsarahm:
Thanks!

Issue #1105, opened 2020-04-13
Include Core Team on Carpentries Roles worksheet


maneesha:
I think this will help new community members understand what the Core Team does.

karenword:
This worksheet needs updating in several ways now (Community Facilitators!) and it we could also use a fillable version.

FWIW I tried implementing a smaller version of it as a google form in my last training and... it kind of worked? Probably not an optimal solution but:
https://docs.google.com/forms/d/12edxdQ3G28npIBef5KCxjMUjZHbacWAdtGzTvcbU5Qc/edit?usp=sharing

Denubis:
Is it worth me making a LaTeX version that's more maintainable?

brownsarahm:
I would like that.

mikemhenry:
Feedback facilitators too!

Denubis:
Ok, let me see what I can do

Denubis:
@mikemhenry What do they do in 15 words or less?

Denubis:
Quick proof of concept

https://www.overleaf.com/read/svyjzwrpgmnm
[Carpentries_Roles_Quiz.pdf](https://github.com/carpentries/instructor-training/files/5551482/Carpentries_Roles_Quiz.pdf)

brownsarahm:
maybe un-nest the last two to make it space to answer them separately?

Denubis:
[Carpentries_Roles_Quiz (1).pdf](https://github.com/carpentries/instructor-training/files/5557343/Carpentries_Roles_Quiz.1.pdf)

@brownsarahm happy to share edit access of the TeX. I've tried to keep the file trivial so it can be maintained by everyone, but holy expert awareness gap, batman.

brownsarahm:
Would it make sense to add it to this repo for version control here? 

(though then the tex environment needs to be managed, etc)

Denubis:
So from a "technically correct" point of view, we can add it as a submodule to this repo -- but since submodules are even more tricky to get right and non-github ones don't have the github affordances it would mostly be technical performance.

I think the first question is: is the `main.tex` file sufficiently legible to be editable causally? the problem is that I could probably pandocise this whole thing, but that just adds another layer to the tower of complexity. I know R has a tinyTeX compiler, so maybe we should head in that direction?

Denubis:
[...]

Issue #1173, opened 2020-09-23
Repository-specific styles & data (resource for contributors)


karenword:
This issue will serve as a storage location for stylistic data used in this repository, e.g. figure fonts, color palettes etc. until/unless a better location presents itself. 

The [Carpentries Venn Diagram](https://github.com/carpentries/instructor-training/blob/gh-pages/fig/carpentries-venn-diagram_20200904.svg):
- Font name is Adobe Courier 
- font size is 8pt and it is bold

Carpentries brand colors (not consistently used in this repository but handy for new contributions) may be found here: https://docs.carpentries.org/topic_folders/communications/resources/brand_identity.html?highlight=hexcode

PR #1190, opened 2020-11-23
R code to parse and run TeX quiz without extra files cluttering every…

Denubis:
PR in response to #1105 

Hopefully this R code is sufficiently self contained as to be useful to other people. It is *very much* a hack since R isn't one of "my" languages. It would be interesting to know how this runs for folk who don't have TeX on their computers. The Rnw file should also be "knittable" in RStudio though. @brownsarahm, thoughts about how we should document this?


brownsarahm:
The documentation question, the temporary consensus was to put things like this in #1173 . 

 I requested reviews from the core-team members that I think would have very helpful insight. If we consider this a pilot and the feature might get used in other lessons, then maybe we temporary document it with a in-file comment and then move the docs to lesson-example after some period of time?

Denubis:
Thinking about this this morning -- this same infrastructure can be used to make printable lessons in far higher quality than the AIO printable browser thingo. But this specific tech... thing is to solve the problem of editing an interactive quiz-handout without needing to use word, export to pdf, open in acrobat, drop in form fields, resave, commit...

ragamouf:
Am I right in thinking this solves the problem of how to make an interactive quiz reusable in a practical and reasonably straightforward way? As neither an R or TeX person, I might need some help actually implementing it. I do have a quiz or two that I'd like to mark up in this way though. Then I could draft some documentation to replicate steps. Would that serve as a useful proof of concept part 2?

Denubis:
@ragamouf Yes -- though right now it implements matching and short answer interactivity only.

karenword:
Despite its awesomeness this is not necessary for the current version of the curriculum.

Issue #1189, opened 2020-11-19
Note in demo info that some are "themed"


brownsarahm:
I learned about this while signing up.  I figure this should be documented in the curriculum at least so that trainers mention it/notice when preparing.

ChristinaLK:
are demos themed, or discussion sessions?

brownsarahm:
The signup now includes 

>Would you like to reserve this time slot for any particular community group? If so please note the group here. Note that this will be recommended on the Etherpad signup, but can not be enforced.

and maneesha confirmed that this is for demos

maneesha:
There might be a better way to phrase this other than "themed."
The intent was to say the session was intended for certain groups.
Some examples:

* some of our Australian Trainers wanted to hold seats for Australian trainees
* demo may be conducted in Spanish

ChristinaLK:
AHA, that makes sense. Got it. Agree w/ Sarah that we could add a note to the curriculum.

PR #1194, opened 2020-12-10
note sessions with themes or community focus, closes #1193 #1189

### LINKED ISSUE #1193: Preparation for Community Discussion as part of Checkout
One person commented on a Community Discussion Feedback Form: "The preparation for the discussion that we received in advance (preparing points to talk about a lesson) was not the same as what we did in the call. " 

And it is true that what trainees are asked to prepare on this [Checkout Procedure Page](https://carpentries.github.io/instructor-training/checkout/) is not always the format community discussions take. Although any extra prep with the lessons is good for a new instructor, I can imagine that sometimes people could be frustrated if they prepare questions and comments and then don't get a chance to use them. 
  
Do others think that what we ask trainees to do for the community discussion matches what happens in a community discussion?

brownsarahm:
I think this addresses the two issues mentioned to note that community discussions have different formats and that some demos are focused. 

@maneesha your input on the language re: demos would be helpful


ChristinaLK (commented):
Looks good to me! Thanks @brownsarahm.

karenword:
Thank you @brownsarahm !

PR #1199, opened 2021-01-16
Karenword mindset motivation overhaul

karenword:
This is an initial draft that combines mindset into motivation and separates out accessibility and inclusion.


karenword:
@brownsarahm this is what I've managed to accomplish this week. It's far enough along that you can see where I'm headed, and if you're feeling plucky you could probably go ahead and teach from it since none of the activities are new. I will continue work on this next week.

karenword:
Note that I have made some minor changes to the accessibility content but the diff won't display because this content has been moved. :(

karenword:
Whoops! This ended up in the wrong repo. Closing and trying again in data-lessons

PR #1209, opened 2021-01-28
Update 21-carpentries.md

christadk:
Some minor language style errors fixed in the "How we operate" session as part of the Instructor Training.


karenword:
thanks @ChristinaLK !

PR #1211, opened 2021-02-02
Update broken URL for Greg's post

zkamvar:
It appears that his blog moved from `thing.html` to `thing/index.html`.


karenword:
thanks @zkamvar !

PR #1236, opened 2021-04-01
Update links to HTTPS

Bisaloo:
This is mostly important for the link to the community discussion pad, which currently links to a page that doesn't support HTTPS (before redirecting to another domain that supports HTTPS). Therefore, this causes issues in browsers that are configured for HTTPS-only navigation.


brownsarahm:
Thank you @Bisaloo ! This is an important fix, so I'm going to merge it now, but also open an issue for checking this throughout the curriculum as a part of the current major update.

PR #1240, opened 2021-04-20
Language Tone and Evidence for Claims in Part 03 — Expertise

humanfactors:
This pull request addresses two core issues in the 3rd component (Expertise).

**1) Less stereotypical language of researchers**
The language used to describe researchers inadvertently was quite stereotypical in tone, and made generalisations about a broader group in a tone that did not conform well with the community standards. The minor edits (5e5c21a) address this issue, and also fix a minor grammatical issue around installation.

**2)  Improve section on the expertise-reversal effect**

The existing discussion of the expertise-reversal effect was inaccurate. The effect does not suggest that experts are "less good" at teaching than novices. What it suggests is that the relative amount of instructional support and guidance has differential impact depending on the level of expertise in a topic area. It implies that for experts, walking through all the details of concepts actually can have a negative impact on performance; while for novices, the absence of these materials can have a negative impact.

For example:
"The expertise reversal is a reversal in the relative effectiveness of instructional methods as levels of learner knowledge in a domain change." - https://link.springer.com/article/10.1007/s11251-009-9102-0

https://impact.chartered.college/article/expertise-reversal-effect-instructional-implications/

I've tried to remain faithful to the original points raised, but be more faithful to the scientific body of work underpinning the effect. I have also removed the point that "experts" are poor lecturers, which there is little evidence of, and I do not think supports learners in this environment (and may in fact raise the exact issues with instructors that this article is trying to address).

------

I have not recompiled the repository, however. Any comments on this submission are very much appreciated.

### DIFF
--- a/_episodes/03-expertise.md
+++ b/_episodes/03-expertise.md
@@ -253,9 +253,9 @@ They may not care how hash tables work,
 or even that hash tables exist; they just want to know how to process data faster.
 We therefore have to make sure that everything we teach is useful right away,
 and conversely that we do not teach anything just because it is "fundamental".
-- Believing that something will be hard to learn is a self-fulfilling prophecy. Researchers already believe that computing is hard, a belief that is exacerbated by "experts" telling them that things should be easy when, in the researcher's experience, they are not.  Paradoxically, this is why it’s important not to say that something is easy.  If someone who has been told that tries it, and it doesn’t work, they are more likely to become discouraged.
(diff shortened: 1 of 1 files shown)


humanfactors:
Thanks for this update (sorry for slow reply @karenword). I take it that I can stay put for now, and wait until you resolve issues in the overhaul before this is integrated further? Thanks again.

Issue #527, opened 2017-10-11
home page could use a photo and short summary


raynamharris:
You know how paste a hyperlink to a website on social media, sometimes you get a little pop-up window with a photo and some text from the website? Here are two examples of pop-ups from the instructor training website.

![screen shot 2017-10-11 at 2 34 15 pm](https://user-images.githubusercontent.com/7544197/31462502-963b4866-ae92-11e7-8209-3758a7926dab.png)

![screen shot 2017-10-11 at 2 36 11 pm](https://user-images.githubusercontent.com/7544197/31462505-99e24dca-ae92-11e7-81a2-7d09ecbb1e3d.png)

I think we could improve our home page by adding:

1. a single image with both the SWC and DC logos
2. an opening sentence that states the mission/vision/purpose/objective of instructor training. 

What do you think? Does anyone else agree that we should add these two things

weaverbel:
I agree @raynamharris but won't volunteer to try to add images as that won't end well!

ChristinaLK:
I agree, if we can figure out how to produce the outcome you're looking for.  @raynamharris do you know if we need to tag/flag that material specifically?  Does it go in the header?

fmichonneau:
Resources for facebook are listed here: https://developers.facebook.com/docs/sharing/webmasters and for twitter here: https://developer.twitter.com/en/docs/tweets/optimize-with-cards/guides/getting-started

From a quick glance, it looks like twitter will use the facebook system (OpenGraph) if it doesn't detect twitter specific tags. 

A couple more thoughts:
- this changes should probably affect not only the templates for the lessons (including instructor training) but also the main sites (e.g. software-carpentry.org)
- I think the main site has recently implemented this, but I also think that the image used (the logo) is on transparent background which doesn't look good on twitter dark background for instance.

ChristinaLK:
Thanks @fmichonneau!

raynamharris:
Thanks @fmichonneau! Yes, the old SWC sites used to have the logo on them, so that would appear on facebook and such.

ErinBecker:
I think this is a good idea. I don't know anything about Facebook, however. Could someone put together a PR to implement this?

weaverbel:
I think this is done via the `excerpt` text in the page header @ErinBecker - Jonah and I added excerpt text to all SWC pages so that real text rather than boilerplate text would appear in Twitter grabs.

brownsarahm:
[...]

PR #1246, opened 2021-05-14
OpenGraph details

mjaquiery:
Added OpenGraph meta tags, and a logo image to populate them. Included a Twitter card to display in summary format.

The description is taken from the first available of:
* page.summary
* page.questions
* page.keypoints
* generic site description

The description is truncated to 200 characters.

A .png version of the logo is saved because [Twitter does not support .svg](https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/summary). They do ask us not to use a generic image, but it's useful in our case, I think.

This should help with #527


brownsarahm:
I've requested input from the lesson infrastructure core team members to to see if this conflicts with things from styles or what the best way to handle this is.

mjaquiery:
I wasn’t sure if it would be more useful elsewhere, but whichever way it should serve as a template for changes.

I made it here initially because the `_layouts` templates for lessons and episodes both reference the `base_layouts/.html` template, and that has `<head></head>` matter in it, meaning (to my knowledge of Jekyll) that the sites will use that header content rather than inheriting it from further downfield.

Matt Jaquiery
UKRN Coordinator for Open Research Training
DPhil Candidate (ExPsy), ACC Lab, University of Oxford

zkamvar (commented):
Thank you for bringing this in! It looks like a good solution to the metadata issue (which @tobyhodges and I will be working on better standardizing across lessons later this year). I also like the fallbacks you have for the lesson and workshop descriptions.

The only conflict right now is that **this PR should be in the [carpentries/styles](https://github.com/carpentries/styles) repository** since this metadata is useful across lessons.

Other than that, here are my suggestions about the implementation:

1. The home page shows the schedule with the questions for each episode, but I think it might be better to prioritize keypoints over questions for the metadata since it will often be viewed out of context of the lesson. 
2. I believe we can use [the largest mstile favicon](https://github.com/carpentries/instructor-training/blob/gh-pages/assets/favicons/cp/mstile-310x310.png) instead of uploading a new icon. You can see how these are parsed in [_includes/favicons.html](https://github.com/carpentries/instructor-training/blob/gh-pages/_includes/favicons.html) (and maybe all of this should go in this particular file since it all goes in the `<head>` tag.

fmichonneau (commented):
I agree with @zkamvar's comments. Could you please re-open this against carpentries/styles? Thanks!

brownsarahm:
Thanks for opening it on [styles](https://github.com/carpentries/styles/pull/604) I'm closing this PR.

PR #1250, opened 2021-07-13
etherpad

Denubis:
@karenword Here are the updates we spoke about during our call. Removed callouts, turned headers into bold font, broke out testimonials better. Longer term, getting outlines into the YAML will help with preparation and "speakers notes" and we can use yaml-based datastructures more effectively than pure markdown.


karenword:
thanks @Denubis !

PR #1251, opened 2021-07-16
05-memory.md: Python codeblocks: change .python to .language-python

maxim-belkin:
Use `.language-python` instead of `.python`. The `.language-*` code block language specifiers were added in carpentries/styles@6a3a807 back in 2018.


brownsarahm:
HI @maxim-belkin Thanks for this, but we've removed the python examples from the lesson to be more inclusive to instructors who are not programmers.

PR #1286, opened 2021-12-09
Retitles references page

karenword:
I'm unfamiliar with the page:reference in the yaml header so I don't know for sure that this will work, but it seems like a legitimate way to change the title of the References page to something other than "Books"?

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -1,5 +1,6 @@
 ---
 layout: reference
+title: "Key Points and References"
 root: .
 ---
 


reference.md
```diff
@@ -1,5 +1,6 @@
 ---
 layout: reference
+title: "Key Points and References"
 root: .
 ---
```
brownsarahm, line 6: "Books" comes from the following line as far as I can tell

brownsarahm (commented):
I htink this is worth merging to see (it shoudn't cause any errors) but this behavior isn't what it should be as far as I can tell.

this is a weird interaction, the layout for the page [reference](https://github.com/carpentries/instructor-training/blob/gh-pages/_layouts/reference.html) actually says the title should be "References"  which is then what the included on all pages [main title](https://github.com/carpentries/instructor-training/blob/gh-pages/_includes/main_title.html) should use, but somehow it's getting "Books"

karenword:
maybe @zkamvar would be interested in this weirdness

zkamvar:
I believe this is the correct fix for the problem.

PR #1287, opened 2021-12-15
Changed wording of rubric headings...

arockenberger:
... according to the updated Instructor Training curriculum

### DIFF
--- a/_extras/demos_rubric.md
+++ b/_extras/demos_rubric.md
@@ -11,13 +11,13 @@ In general, the majority of people pass their demonstrations. When a trainee is 
 mark in a "Negative" column can be considered justification for asking a trainee to redo their demonstration if the problem is significant. As always,
 Trainers should use their own judgement when applying this rubric in individual cases.  
 
-|Positive Content|Negative Content|
+|Positive Content|Opportunities for Growth|
 |------|---------------------|
 Uses Carpentries curriculum with only minor deviations|Deviates significantly or does not use Carpentries curriculum|
 Places content in context and explains relevance/utility to learners|Jumps into the content without context|
 Teaches content correctly|Makes factual errors in content and does not correct|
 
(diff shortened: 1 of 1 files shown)


_extras/demos_rubric.md
```diff
@@ -11,13 +11,13 @@ In general, the majority of people pass their demonstrations. When a trainee is
 mark in a "Negative" column can be considered justification for asking a trainee to redo their demonstration if the problem is significant. As always,
 Trainers should use their own judgement when applying this rubric in individual cases.  
 
-|Positive Content|Negative Content|
+|Positive Content|Opportunities for Growth|
```
brownsarahm, line 14:
```suggestion
|Positive Content|Content Opportunities for Growth |
```

_extras/demos_rubric.md
```diff
@@ -11,13 +11,13 @@ In general, the majority of people pass their demonstrations. When a trainee is
...
 |------|---------------------|
 Uses Carpentries curriculum with only minor deviations|Deviates significantly or does not use Carpentries curriculum|
 Places content in context and explains relevance/utility to learners|Jumps into the content without context|
 Teaches content correctly|Makes factual errors in content and does not correct|
 
-|Positive Delivery|Negative Delivery|
+|Positive Delivery|Opportunities for Growth|
```
brownsarahm, line 20:
```suggestion
|Positive Delivery|Delivery Opportunities for Growth|
```

brownsarahm (dismissed):
Good catch @arockenberger ! We should fix this, but I think it needs a minor change

I think "content" and "delivery" should be headings on both columns?

or maybe subheadings over the separate tables instead of in either column?  

Open to feedback on which is better, but I think the lack of parallelism between the columns of the table is not what we want.

karenword:
thanks @arockenberger !

PR #1293, opened 2022-01-03
clarify language re profile creation

### LINKED ISSUE #1291: Confusion on how and when an instructor profile is created
The information in https://carpentries.github.io/instructor-training/checkout/index.html#4-tracking-your-progress states that a profile in the Carpentries database AMY is created **after** the Instructor Training workshop **by someone else**.

On the contrary, the section "There are two ways to access Instructor Training" at https://carpentries.org/become-instructor/ mentions that a profile is created **by the candidates** themselves **before** the workshop, when they sign up. "When you register, you will also be asked to create a profile in our database".

At this point it is not clear which procedure applies, whether the two statements refer to the same profile in the first place, and how one should interpret it when the access to the Carpentries database is denied (that is, is the failure expected or unexpected?).

karenword:
Fixes #1291

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -247,7 +247,7 @@ also email you to let you know, as this information is typically not shared duri
 after you have attended your teaching demonstration.
 
 ## 4. Tracking Your Progress
-After your instructor training workshop workshop, Carpentries Core Team members will activate
+After your instructor training workshop workshop, Carpentries Core Team members will use your application or profile-creation form to activate
 your profile in The
 Carpentries database, AMY. You can monitor your checkout progress by [logging on to our database, AMY][amy-login], using your GitHub ID.
 If you have any questions or did not provide a GitHub user name in your form, please email us 
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -247,10 +247,11 @@ also email you to let you know, as this information is typically not shared duri
 after you have attended your teaching demonstration.
 
 ## 4. Tracking Your Progress
-After you complete this workshop, your application will be matched to your training record and a profile will be created in The
+After you complete this workshop, your application or profile-creation form will be matched to your training record and used to 
+create a profile that you can access in The
 Carpentries database, AMY. You can monitor your checkout progress by [logging on to our database, AMY][amy-login], using your GitHub ID.
-If you have any questions or did not provide a GitHub user name when registering for
-training, please email us at [instructor.training@carpentries.org](mailto: instructor.training@carpentries.org).
+If you have any questions or did not provide a GitHub user name in your form, please email us 
+at [instructor.training@carpentries.org](mailto: instructor.training@carpentries.org).
 
 Keeping [your profile][trainee-profile] up to date with a current email address and local airport helps us to keep in touch about teaching
 opportunities. You may also adjust your preferences to select whether you want to have your profile shared on our
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -247,8 +247,8 @@ also email you to let you know, as this information is typically not shared duri
 after you have attended your teaching demonstration.
 
 ## 4. Tracking Your Progress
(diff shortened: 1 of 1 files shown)


_extras/checkout.md
```diff
@@ -247,10 +247,11 @@ also email you to let you know, as this information is typically not shared duri
 after you have attended your teaching demonstration.
 
 ## 4. Tracking Your Progress
-After you complete this workshop, your application will be matched to your training record and a profile will be created in The
+After you complete this workshop, your application or profile-creation form will be matched to your training record and used to 
+create a profile that you can access in The
```
brownsarahm, line 251: this is hard to parse.  doe s it work like:
1. application or profile creations form at time of registration
2. workshop
3. training record added so that progress can be monitored

modificaitons can be made (and github linked) by e-mailing

karenword: yes. Basically AMY can take an application and turn it into a "person" profile where training records are tracked, but this requires us to press buttons, which we don't do until after someone has attended an IT event. Would it help to have numbered steps here?

_extras/checkout.md
```diff
@@ -247,10 +247,11 @@ also email you to let you know, as this information is typically not shared duri
...
 Carpentries database, AMY. You can monitor your checkout progress by [logging on to our database, AMY][amy-login], using your GitHub ID.
-If you have any questions or did not provide a GitHub user name when registering for
-training, please email us at [instructor.training@carpentries.org](mailto: instructor.training@carpentries.org).
+If you have any questions or did not provide a GitHub user name in your form, please email us 
+at [instructor.training@carpentries.org](mailto: instructor.training@carpentries.org).
 
 Keeping [your profile][trainee-profile] up to date with a current email address and local airport helps us to keep in touch about teaching
```
brownsarahm, line 256: also, I don't see `trainee-profile` defined below?

_extras/checkout.md
brownsarahm, line 251:
```suggestion
<X time> after your instructor training workshop workshop, Carpentries Core Team members will activate
your profile in The
```
maybe simplify it to just what impacts the trainee instead of what happens a little bit?  

Do trainees get notified by AMY when they can access it?

_extras/checkout.md
karenword: it's there I think I'm just bad at alphabetizing :)

_extras/checkout.md
```diff
@@ -247,10 +247,11 @@ also email you to let you know, as this information is typically not shared duri
 after you have attended your teaching demonstration.
 
 ## 4. Tracking Your Progress
-After you complete this workshop, your application will be matched to your training record and a profile will be created in The
+After your instructor training workshop workshop, Carpentries Core Team members will activate
```
karenword, line 250:
```suggestion
After your instructor training workshop workshop, Carpentries Core Team members will use your application or profile-creation form to activate
```

karenword: I'm making the use of the form more explicit because failure to submit the form is often a sticking point, so it's good to have one more place to remind them that it is necessary.

brownsarahm: is it choose once application or profile creation form, or is that two names for the same form. 
If someone has not done this, how would they know? how can a trainer help them if they ask questions about that in a workshop?

_extras/checkout.md
brownsarahm: and i'm bad clicking to view the whole page in PR view.

_extras/checkout.md
```diff
@@ -247,10 +247,11 @@ also email you to let you know, as this information is typically not shared duri
 after you have attended your teaching demonstration.
 
 ## 4. Tracking Your Progress
-After you complete this workshop, your application will be matched to your training record and a profile will be created in The
+After your instructor training workshop workshop, Carpentries Core Team members will use your application or profile-creation form to activate
```
cknuepfer, line 250: There is a double "workshop workshop".

PR #1298, opened 2022-01-19
update link and description to aggregate survey results

fmichonneau:


karenword:
thanks @fmichonneau !

PR #1299, opened 2022-01-20
adjust wording of "leave no learner behind" tip

tobyhodges:
Because the info about how to use sticky notes to gauge progress is no longer located in the same episode as the list of tips.


brownsarahm:
Thanks @tobyhodges

PR #1308, opened 2022-02-10
Add link for bonus modules to checkout episode

jcoliver:
The link for bonus modules was missing from the episode about the checkout process (Ep. 14), so I copied the one in /_extras/checkout.md and pasted it into the checkout episode.


brownsarahm:
Thanks @jcoliver

PR #1311, opened 2022-02-14
fix broken link

zkamvar:
The fourth point of the first list in the homepage goes to an invalid link: `https://carpentries.github.io/instructor-training/[Carpentries](https://carpentries.org/)`

This PR fixes that.


brownsarahm:
thanks @zkamvar !

PR #1318, opened 2022-03-01
Small formatting updates to episode 02-practice-learning

anenadic:
Minor formatting suggestions:
- consistent formatting of lists in exercises (the use of `.` vs. `)` consistently)
- removed a superfluous blank space
- added a blank space to separate "how much time is needed sentence" in exercises from the rest of the text, mainly for consistency but also to make it stand out more

### DIFF
--- a/_episodes/02-practice-learning.md
+++ b/_episodes/02-practice-learning.md
@@ -115,7 +115,6 @@ If you choose to use an analogy, be sure its usefulness outweighs its potential 
 > you prefer to work with an example, consider this common analogy from education: "teaching is like gardening."
 > 2. Share your analogy with a partner or group. (If you have not yet done so, be sure to take a moment to introduce yourself, first!) What does your analogy
 > convey about the topic? How is it useful? In what ways is it wrong?
-> 
 >
 > This activity should take about 10 minutes.
 {: .challenge}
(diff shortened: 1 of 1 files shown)


_episodes/02-practice-learning.md
```diff
@@ -112,10 +112,9 @@ If you choose to use an analogy, be sure its usefulness outweighs its potential
 > ## Analogy Brainstorm
 >
 > 1. Think of an analogy to explore. Perhaps you have a favorite that relates to your area of professional interest, or a hobby. If 
-> you prefer to work with an example, consider this common analogy from education: "teaching is like gardening."
+> you prefer to work with an example, consider this common analogy from education: "teaching is like gardening".
```
brownsarahm, line 115: based on the [style guide](https://docs.carpentries.org/topic_folders/communications/resources/style-guide.html#quotation-marks) I'm actually not acturally sure we should move this outside, since "teaching is like gardening." is a full sentence

_episodes/02-practice-learning.md
```diff
@@ -173,19 +172,18 @@ perhaps illustrated as:
...
 > What challenges might a novice face in creating a concept map of this kind?  
+> 
 > This exercise should take about 5 minutes.
 {: .challenge}
 
 ## Misconceptions
 
-The mental model above connects a ball to the water it can displace, recognizing that 'more' ball can move 'more' water. This mental model is perfectly functional for a child who wants to have fun splashing water around.  
-It may endure in this way for several years of beaches and bathtubs.
+The mental model above connects a ball to the water it can displace, recognizing that 'more' ball can move 'more' water. This mental model is perfectly functional for a child who wants to have fun splashing water around. It may endure in this way for several years of beaches and bathtubs.
```
brownsarahm, line 186: I know we're not very consistent at this overall, but it's actually better for version control to keep that newline so that sentences are separate lines, it makes a diff easier to read.

brownsarahm (commented):
Thanks @anenadic for the careful read, I think some of these deserve some discussion though.

_episodes/02-practice-learning.md
anenadic: Yes, I see you point @brownsarahm - in which case we might need two fullstops to end the outer sentence? 
"'.. you prefer to work with an example, consider this common analogy from education: "teaching is like gardening.".'

_episodes/02-practice-learning.md
anenadic: If that sentence is part of the previous paragraph - we should probably remove the new line. The sentence before it also does not start on the new line so I do not see why this sentence is special.

_episodes/02-practice-learning.md
brownsarahm: I don't think we put two either. the one full stop does double work in this case.

_episodes/02-practice-learning.md
brownsarahm: I know that we're not consistent, my intentions was that the better choice, for version control is to add an additional new line in the markdown.  Newlines are not paragraph breaks, blank lines are paragraph breaks. I work in more strict environments a lot though, so here I'm okay to defer to @karenword

anenadic: I am attaching the screenshot here to show what the problem is - I think we have 2 blank spaces which causes the sentence to start on a new line, rather than continue in the same paragraph. I agree that in source Markdown having sentences start on a new line is handy for easier diffs - but this is a separate issue I believe. Took me a while to figure out what is wrong and that @brownsarahm and I were talking about slightly different issues.
Screenshot 2022-03-04 at 16 34 01

anenadic: I've done the easy fix in PR #1327.

brownsarahm: ohhh sorry, i had not rendered, i was looking at the source and white spaces at the end of the line also make a new paragraph.

brownsarahm: thank you for your patience and diligence!

PR #1319, opened 2022-03-01
Align text to images in 04-expertise

anenadic:
A few suggested fixes for episode 04-expertise:

1. fixed steps of reasoning for a competent practitioner (around metal models of people with different skills)

The steps described in the text do not match the diagram (of the mental models of novices/competent practitioners and experts) - we either need to change the diagram or the explanation of it. I've changed the explanation but it may be better to modify the diagram to swap nodes E and F in the diagram.

This was issue with the old curriculum as well, but I never got round to fixing it.

2. list consistency improvement


karenword:
Thanks @anenadic this was bothering me just last week and I never stopped to recognize that it was something we could fix!

PR #1316, opened 2022-02-24
Various small corrections

fiveop:
Each (type of) change has its own commit, so they can be easily be removed, if reviewers disagree.

Each change adds, removes, or changes at most a handful of words.

PR #1324, opened 2022-03-03
remove obsolete trainer note

tobyhodges:
This note refers to the Flash-based tool we used to use for the working memory challenge, and should be removed for the avoidance of confusion.


brownsarahm:
this is also done by #1316 which includes many other changes, and jsut needs confirmation of one small change before merging

tobyhodges:
Cool. I will close this now. Looks like your suggestion in #1316 was addressed earlier today, @brownsarahm ?

brownsarahm:
Yes, just before I saw this...

PR #1329, opened 2022-03-06
fix: modify time slot in training calendar

amangoel185:
Fix the training time-slot on [March 24-25, 2022](https://carpentries.github.io/instructor-training/training_calendar/index.html#march-24-25-2022) to `2 pm - 10 pm` instead of `2 pm - 8 pm`


maneesha:
Thanks for catching this @amangoel185 !

PR #1331, opened 2022-03-07
Remove extra verb

fiveop:
As an alternative to the proposed change, we could keep the other verb or both (connected by an and).


brownsarahm:
thanks!

PR #1339, opened 2022-03-16
make link text more accessible

tobyhodges:
Expanding the link text to make the page more accessible e.g. for visitors using a screen reader.


brownsarahm:
thanks Toby! this is a very thoughtful change 😍

PR #1340, opened 2022-03-17
Move minute card from before break to before lunch

### LINKED ISSUE #1307: Move minute card prompt from episode 21 to 18
A prompt for learners to give minute card feedback comes at the end of Episode 21 Working With Your Team, right before a(n afternoon) break, but it seems it should instead be at the end of Episode 18 Preparing to Teach, which is the episode immediately before the lunch/half-day end. :thinking:

brownsarahm:
closes #1307


karenword:
thanks @brownsarahm and @jcoliver !

PR #1346, opened 2022-04-01
Adding in plotting for python inflammation lesson

sstevens2:
This episode starts with the data already loaded so it is a hard place for someone to start the teaching demo.


brownsarahm:
I will merge this if the maintainers of the lesson reject [my PR to reload the data at the start of the episode](https://github.com/swcarpentry/python-novice-inflammation/pull/986)

I think, however, there's multiple benefits to re-loading the data. This is  minimal dependency because the prior episode does not modify the data at all, so it's a one line fix to to the episode to make this suitable and more starting points is helpful when multiple trainees pick the same lesson.

sstevens2:
Thanks @brownsarahm!  I think that is good alternative fix.

PR #1360, opened 2022-04-26
TD Rubric Template

fiveop:
The first commit changes the rubric form into a code block in markdown format for easy copying into text editors. The way it was, I did not see an easy way to use it (If there is, it could be mentioned instead of this change).

The second commit adds “Manual typing” to that form, the only point that was missing relative to the table above.

### DIFF
--- a/_extras/demos_rubric.md
+++ b/_extras/demos_rubric.md
@@ -33,24 +33,30 @@ Appears focused and attentive to learners|Appears significantly distracted by co
 
 ---
 
-Trainers can also use the form below to take notes during the teaching demonstration.
+Trainers can also copy the form below to take notes during the teaching demonstration.
 
-Trainee:
+```
+# Teaching Demonstration Notes
 
-Date:
+**Trainee**:
+**Date**:
 
-|Aspect|Comments_________________________________________________|
-|------|---------------------|
-|<br>Sticks to curriculum<br>|
-|<br>Context|
-|<br>Content Knowledge|
-|<br>Mirror learner's environment|
-|<br>Font & window size|
-|<br>Explanation of typing|
-|<br>Embraces & uses mistakes|
-|<br>Use of notes on paper/tablet|
-|<br>Distraction-free screen|
-|<br>Wording|
-|<br>Pacing|
-|<br>Clarity|
-|<br>Demeanor & focus|
+## Content
+
+**Sticks to curriculum**:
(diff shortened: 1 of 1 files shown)


brownsarahm:
This has potential to significantly impact how trainers operationally conduct their teaching demos, so I am going to solicit broader feedback on it.

brownsarahm:
This change would not change what trainers are *required* to do, just how this page supports them in making evaluations.  No trainer would be required to use this exactly.

ha0ye:
Ah, sorry, per @brownsarahm's nudge, I see that the PR is for the form at the bottom of the teaching demo rubric page. I cannot recall ever having scrolled down to that section of the page before!

So a more accurate framing of my comment is probably, this change doesn't bother me, but I'd rather the form/checklist did not exist at all.

JCSzamosi:
fwiw, I really like having the checklist. It doesn't fully determine whether I pass someone or not, but it is extremely helpful to me to have a concrete reminder in front of me during demos for what kinds of things I'm looking for. I certainly don't think anyone should be required to use it, but it would make my life worse if it didn't exist.

zpainter:
I like having the checklist, even if I don't strictly stick to it. It feels like one of the more common grading formats, that I can point to with other trainers who do these.

fiveop:
I would not use it my self, but when I saw it, I had no idea how it could be used (There is no new content with respect to the table above).

I changed the format, so that it is easily copied. That way I know how it could be useful.

I would not mind removing it entirely as well, but @JCSzamosi  and @zpainter already mentioned that they use it. Maybe you could tell us how and if the change would be an improvement.

JCSzamosi:
> I would not mind removing it entirely as well, but @JCSzamosi and @zpainter already mentioned that they use it. Maybe you could tell us how and if the change would be an improvement.

I'm not sure I understand the question. It's basically a checklist of carpentries teaching practices we want to see in trainees, right? So as they do their demo, I check off the things they do well. It isn't the whole basis of my decision, but it helps me remember what I'm looking for.

As for whether the change will be an improvement, I haven't been able to figure out how to view a rendered version of the PR so I can't say for sure, but from the description (sticks the list in a code block) I don't see how it could possibly make it worse, so I think it's fine.

brownsarahm:
I'm going to merge this since there was no major objections to the *format* changes proposed here.

PR #1373, opened 2022-05-31
Reorder solutions

karenword:
This PR moves the suggested item up in the list to be grouped with other follow-up type things and leave "celebrate" for last. It also adds a bullet for this item. I'm not sure that the list will actually render sub-bullets, so it may be desirable to strip some of the detail here. However, I'm not too concerned at the moment because I think many of these "solutions" boxes will end up being re-done with in-line Instructor Notes in the workbench.


brownsarahm:
Do we want to create an issue as a flag to come back to re-ordering in this location? Otherwise this looks good.

PR #1383, opened 2022-06-13
Remove item from list

fiveop:
The instructions state to copy the objective and then add it as the first entry of a list right below. That is redundant.


brownsarahm:
thanks!

PR #1387, opened 2022-06-15
Make time references to 4 parts, not 2 days

### LINKED ISSUE #1385: Rewrite temporal references to 4 parts instead of two days
we currently have references to "tomorrow afternoon" and tomorrow morning that could be rephrased to part 3, part 4. This will accommodate longer workshops better.

brownsarahm:
closes #1385 

I did not change episode names to remove "lunch" to avoid breaking links, but maybe we can do that too? but I think I got all of the other temporal references changed to parts instead of two days.

### DIFF
--- a/_episodes/06-feedback.md
+++ b/_episodes/06-feedback.md
@@ -69,7 +69,7 @@ opportunity to reflect), yourself (you will get more useful feedback), and The C
 
 ## Minute Cards
 
-During any lunch breaks and between days, we
+Before each long break, for example lunch or between days, we
 have learners complete **minute cards** to share anonymous feedback. At an in-person workshop,
 paper sticky notes double as minute cards, with the two different colors used for positive and
 constructive feedback. At an online workshop, this may be done by making a copy of our [Virtual Minute Card Template](https://docs.google.com/forms/d/1p7iOV5HNvy4POS4g6eottY8RSfKq4kaoKz1-jIFYTMI/template/preview)
--- a/_episodes/06-feedback.md
+++ b/_episodes/06-feedback.md
@@ -60,7 +60,7 @@ a dashboard with the results! Take care not to share this link with your learner
 
 We have found that learners are much more likely to fill out the post-workshop survey
 while they are still at the workshop than they are after they leave the venue. At the end
-of a two-day workshop, your learners' brains will be very tired. Rather than trying to
+of a multi-day workshop, your learners' brains will be very tired. Rather than trying to
 fit in another 15 minutes worth of teaching, give your learners time to complete the
 post-workshop survey at the end of your workshop. You will be helping them (with an
 opportunity to reflect), yourself (you will get more useful feedback), and The Carpentries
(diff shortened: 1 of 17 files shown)


ndporter:
Made minor updates to feedback section. Link 62 was one reference to two-day workshops changed. Lines 72 and 91, I think I solved the lunch problem within the episode - it was saying "at lunch and after day 1" when what we really meant was any time there's a lunch or day break. Hopefully I edited the PR correctly - have not done edits on an existing PR from someone else before.

ndporter:
Re: changing episode names from "xx-lunch": It looks like there aren't any dependencies in the raw episode code for other episodes on those names (since moving between episodes is automatic). The only question I don't know is how the schedule is created on the index page. Ideally, that should actually be split into four half-day chunks as well, but even if that isn't done - that's the only place I couldn't verify whether changing episode names would impact things.

ndporter:
> these updates are about the change from two to four day _workshops_ not instructor training, so I was leaving those references out of this PR intentionally. I did not look at any references to the format of technical workshops elsewhere, only the within IT temporal references.

@brownsarahm should I peel those off into a different PR then? Or merge them all in here?

I assumed that since it's the same issue (both IT and TW can now be taught as 4-half-day instead of 2-whole-day) it didn't make sense not to change those here. And clearly, the technical workshops don't always have lunches, so they do need changed somewhere along the way.

_episodes/06-feedback.md
```diff
@@ -57,44 +57,44 @@ a dashboard with the results! Take care not to share this link with your learner
...
+opportunity to reflect), yourself (you will get more useful feedback), and The Carpentries
 (we improve our programs based on feedback, and our funders take pride in the success of our workshops, too).
 
 
 ## Minute Cards
 
-Before each lunch break and at the end of the first day, we
-have learners complete **minute cards** to share anonymous feedback. At an in-person workshop, 
-paper sticky notes double as minute cards, with the two different colors used for positive and 
+During any lunch breaks and between days, we
```
brownsarahm, line 72:
```suggestion
Before each long break, for example lunch or between days, we
```
I mostly want to keep the "before" not "during" while it may in practice sometimes bleed into break time, we should at least try to advise people to do this before the break time starts.

_episodes/06-feedback.md
```diff
@@ -57,44 +57,44 @@ a dashboard with the results! Take care not to share this link with your learner
...
 **Example constructive prompts:**  
 - One thing you did not like or would change about this section of the workshop
 - One thing that is confusing / you would like clarification on.    
-- One question you have 
+- One question you have
 
-Over lunch and before the second day starts, the Instructors read through the minute 
-cards and look for patterns. At the start of day two and each afternoon, the Instructors
-take a few minutes to address commonly raised issues with the whole class. The 
+During any lunch breaks and between days, the Instructors read through the minute
```
brownsarahm, line 91:
```suggestion
During long breaks, instructors read through the minute
```

brownsarahm (commented):
I guess it does make sense to address these too, but I was more trying to say that I have not checked in case there are other instances.

PR #1391, opened 2022-06-20
An improvement to explaining 'participatory segment'

anenadic:


setup.md
```diff
@@ -15,7 +15,7 @@ the [Data Carpentry lessons page]({{ site.dc_site}}/lessons/), or the
 (for recommendations, see the list below) and read through it carefully. 
 You will be asked to teach a 90 second introduction to this content during 
 the first half of the workshop, and will teach a 3 minute participatory segment 
-of its content twice during the second half of the workshop. Note that these 
+of its content twice during the second half of the workshop (e.g. using live coding or live teaching demonstration). Note that these
```
brownsarahm, line 18:
```suggestion
of its content twice during the second half of the workshop (e.g. using live coding or live demonstration). Note that these 
```
I think these are both teaching, so it's more clear to not use teaching in the second one?  


minor, I think this is an exhaustive list so then "i.e." (that is) is more appropriate than "e.g." (for example).

brownsarahm (commented):
minor: I think this is an exhaustive list defining "participatory segment" not a set of examples from a larger set of options, so then "i.e." (that is) is more appropriate than "e.g." (for example), but I'm open to reasons why it should be "e.g."

setup.md
brownsarahm: @anenadic is this change acceptable to you, or would you like to recommend further changes?

karenword:
thanks @anenadic!

Issue #1395, opened 2022-06-22
Improve accessibility of Instructor Training for people with ASD


sjmf:
This issue regards improving the accessibility of the instructor training for people with autism spectrum disorder and other learning and attention disabilities (e.g. ADHD), in line with the Carpentries Core Value of Access for All. I have a small number of suggestions– not a comprehensive list by any means– of ways in which the accessibility of the content could be improved. These are intended as an __introduction__ to adjustments for the instructor training, and the start of a conversation about inclusion: not a complete guide.

I will post the detail of this issue as individual comments for accessibility purposes: please see below for further information.

Contents:

  0. [Background on Autism Spectrum Disorder](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163277702)
  1. [Invite participants to disclose a disability](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163277885)
  2. [Clarify the cancellations and attendance policy for disabled participants](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278095)
  3. [Consider mental fatigue when scheduling workshops](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278279)
  4. [Reduce visual noise and task switching](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278573)
  5. [Improve information presentation for working memory](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278798)
  6. [Useful adaptations - things that went well](https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278874)

sjmf:
## 1. Invite participants to disclose a disability

We can invite participants to disclose any disabilities which may impair their ability to participate in the training as part of the the sign-up process, or as a part of the welcome email. Inviting participants to declare a disability will help both the individual and the instructors to make adjustments for better inclusion of that person. For example, by adding a line to the training invitation email similar to the following: 

> If you have a disability which might prevent you from being fully included in the instructor training workshops, and you would like us to know about it in advance, please let us know by emailing your trainers directly, or by contacting [instructor.training@carpentries.org](mailto:instructor.training@carpentries.org).

Instructors can then have a conversation with an individual with regards to specific adaptations that can be made to the scheduled training to better include them. This goes beyond autism and is good practice for people with other disabilities (both visible and invisible).

sjmf:
## 4. Visual noise and Task Switching (Etherpads)

[...]

PR #1405, opened 2022-06-23
add an option to the scheduling exercise

### LINKED ISSUE #1404: clarify that signups can be completed later
## 4. Visual noise and Task Switching (Etherpads)

I found the Etherpad useful during the lesson exercises, when participants were populating content into the page shared between the group. However, the instructor checkout process requires signing up for a community discussion and a teaching demo using the pads at https://pad.carpentries.org/community-discussions and https://pad.carpentries.org/teaching-demos. 

The information density of these two pages can be stressful when just five minutes is allocated for signing up within the training schedule. Being presented with this page (with no indexes or hyperlinks to find specific sections within it!) is a high amount of visual noise and it is difficult to navigate, especially when attention is also split with the Zoom call. Philipp handled this really well and clarified that I could come back to these later, but this could be emphasised within the teaching material, or perhaps a more streamlined way could be found to track sign ups to these sessions.

This could also be due to difficulties task switching: changing from a lecture-based learning approach to having to interpret large amounts of data.

_Originally posted by @sjmf in https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278573_

add an option in the [related exercise](https://carpentries.github.io/instructor-training/14-checkout/index.html#schedule-a-discussion-or-demo) to save the links to sign up/ add them to a todo list.

brownsarahm:
closes #1404 per the great advice from #1395

### DIFF
--- a/_episodes/14-checkout.md
+++ b/_episodes/14-checkout.md
@@ -50,8 +50,10 @@ Extensions may be granted for any reason up to 1 year from your training date.
 > If you would prefer to do your teaching demonstration before your discussion,
 > visit the demo Etherpad and sign up there: <https://pad.carpentries.org/teaching-demos>
 > This demo [rubric]({{ page.root }}/demos_rubric/) is provided as a guide for Trainers evaluating potential new Instructors during the teaching demonstration.
->
->
+> 
+> If these pages are hard to navigate or you cannot find a suitable time, save these links where 
+> you can find them again and make a note to schedule these steps of your checkout process. 
+> 
 > This exercise should take 5 minutes.
 {: .challenge}
 
--- a/_episodes/14-checkout.md
+++ b/_episodes/14-checkout.md
@@ -52,7 +52,7 @@ Extensions may be granted for any reason up to 1 year from your training date.
 > This demo [rubric]({{ page.root }}/demos_rubric/) is provided as a guide for Trainers evaluating potential new Instructors during the teaching demonstration.
 > 
 > If these pages are hard to navigate or you cannot find a suitable time, save these links where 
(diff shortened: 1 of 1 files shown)


brownsarahm:
reverted to draft based on [advice](https://github.com/carpentries/instructor-training/issues/1404#issuecomment-1164001848) from @fiveop to provide context here for trainers.  I would like help phrasing that.

brownsarahm:
Yes, that is the right section @brynnelliott I would also appreciate advice on how/if we should put add context for why this option is here to help trainers pass on that context while teaching this section. 

The linked issue #1395 is where we got the idea, so I want to give trainers the right amount of context to help people, but I also don't want to do things that draw unnecessary attention to how this could be an accommodation.

_episodes/14-checkout.md
```diff
@@ -50,8 +50,10 @@ Extensions may be granted for any reason up to 1 year from your training date.
 > If you would prefer to do your teaching demonstration before your discussion,
 > visit the demo Etherpad and sign up there: <https://pad.carpentries.org/teaching-demos>
 > This demo [rubric]({{ page.root }}/demos_rubric/) is provided as a guide for Trainers evaluating potential new Instructors during the teaching demonstration.
->
->
+> 
+> If these pages are hard to navigate or you cannot find a suitable time, save these links where 
+> you can find them again and make a note to schedule these steps of your checkout process.
```
brownsarahm, line 55:
```suggestion
> you can find them again and make a note to schedule these steps for your checkout process. 
```

brynnelliott:
@brownsarahm I think what you wrote gives enough information, this way if someone doesn't want to disclose other details, then they don't have too.

PR #1423, opened 2022-08-22
Refer to handbook for AMY information

### LINKED ISSUE #1354: Refer to Handbook for AMY instructions
We keep instructions for how Instructors can manage their AMY profile in our [Handbook](https://docs.carpentries.org/topic_folders/for_instructors/current_instructors.html#accessing-and-updating-your-instructor-profile).  This should be considered the canonical source of information on how to use AMY.  The [Tracking your Progress](https://carpentries.github.io/instructor-training/checkout/index.html#4-tracking-your-progress) section of the curriculum should refer to this, and not duplicate any information in the Handbook.

ndporter:
Updates information on AMY setup and login to refer to handbook page, reducing need for duplicate maintenance. Addresses and closes #1354

### DIFF
--- a/_extras/checkout.md
+++ b/_extras/checkout.md
@@ -250,9 +250,7 @@ after you have attended your teaching demonstration.
 ## 4. Tracking Your Progress
 After your instructor training workshop, Carpentries Core Team members will use your application or profile-creation form to activate
 your profile in The
-Carpentries database, AMY. You can monitor your checkout progress by [logging on to our database, AMY][amy-login], using your GitHub username.
-If you have any questions or did not provide a GitHub username in your form, please email us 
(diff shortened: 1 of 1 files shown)


brownsarahm:
@karenword @maneesha  the net impact of this change is to direct trainees to team@carpentries.org instead of instructor.training@carpentries.org if they cannot log in is that going to impact core team workflow adversely?

karenword:
Thanks for noticing that @brownsarahm ! I think this should be fine, as this is a problem that the folks on team@ can solve even though it relates to Instructor checkout. If @maneesha disagrees (she is currently on vacation) or we have any problems down the road, we can always update the Handbook.

brownsarahm:
sounds good, I'll merge this now then

PR #1437, opened 2022-09-26
unified badge language

karenword:
(Also removes reference to bonus modules as these are not currently being offered.)


karenword:
hold until @maneesha approves for merge

PR #1441, opened 2022-09-27
Move co-teaching section

martinosorb:
As it is now, the co-teaching section is in an awkward position between a section on feedback and an example of sticky situation. This does not make much sense. When teaching this module recently, I moved it to the beginning, precisely in the place where we are already talking about not teaching alone. This worked for me.

### DIFF
--- a/_episodes/21-management.md
+++ b/_episodes/21-management.md
@@ -32,23 +32,6 @@ In this episode, we will discuss examples, strategies, and resources for handlin
 also take some time to discuss planning logistics, and (time permitting) everyone will have an opportunity to try out
 our instructions for setting up a workshop website.
 
-### Co-Teaching Models
-
-*Co-teaching* describes any situation in which two teachers work
-together in the same classroom. There are [many ways][worldcat-interactions] to do
-this; we use the following methods most frequently:
-
-- **Team teaching:** Both teachers deliver a single stream of content
-  in tandem, taking turns the way that musicians taking solos would.
-
-- **Teach and assist:** Teacher A teaches while Teacher B moves around
-  the classroom to help learners.
-
-"Teach and assist" is most common, in part because people are accustomed to teaching alone! However, 
-"Team teaching" can be a fun way to lead together as well. Team teaching can be particularly helpful when two 
-Instructors have very different strengths, creating a more balanced experience overall. This may also be worth 
-considering for self-organised workshops if a co-Instructor has not yet been certified.
-
 ## The Instructional Team
 A typical Carpentries workshop includes 3 roles; sometimes individuals may choose to play more than one role. Each workshop should have:
 - A **Host** who organizes the workshop logistics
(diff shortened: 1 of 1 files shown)


brownsarahm (changes requested):
This makes the level 3 section alone outside of its level 2 section (The instructional Team). 

It would make more sense to me to put it before  `### Hosting` but still inside `## The Instructional Team`

martinosorb:
It's not alone, it's part of the level 2 section "Never teach alone", which is what this is about.

brownsarahm:
sorry, GitHub's hiding sections in PR view got the best of me here.  
I agree that it should not be after `### Carpentries Classroom Practices`, but  the intro to the `## The instructional team` suggests that this is supposed to be in that section, to create more parallelism to the intro bullet list, maybe this should go right above the `### Helpers` heading? 

I'll make a separate issue to see why `### Carpentries Classroom Practices` is a level 3 at all

brownsarahm:
also thanks for your first contribution to the instructor training curriculum! 🎉

martinosorb:
Ok I have an idea that makes it fit better: the section lists Host, Helpers, Instructors. Then there are sections for Helpers and Host, but not Instructors. The co-teaching section can be the one related to the other instructors, so it could go where you suggested (before helpers), but we could call it "Other instructors" or "Instructors" or "Other instructors: co-teaching" or something? So that the schematic of the three bullet points remains?

martinosorb:
I have updated accordingly

_episodes/21-management.md
```diff
@@ -50,6 +50,23 @@ If you cannot find someone else to bear full responsibility for the Host role, c
 involving helpers in advance of the workshop, or exploring institutional support for event logistics like registration, snack delivery, 
 helper coordination, and emergency contact.
 
+### Co-teaching with other Instructors
```
brownsarahm, line 53:
```suggestion
### Instructors: Models for Co-teaching 
```
I agree with your point that the previous title broke the parallelism and I also liked the term `models` in to emphasize that it's multiple examples, not prescriptive.  What about this?

martinosorb:
Looks good!

_episodes/21-management.md
brownsarahm, line 53:
```suggestion
### Instructors: Co-teaching Models
```

brownsarahm: Since @martinosorb  resolved without commiting, I cannot commit the original.  but the text reply seemed positive, so I've added this again.  If you do *not* want this, resolve again, that's okay.  If you are okay either find the "commit suggestion" button or reply without resolving.

martinosorb:
Sorry, I never used the suggestions before. Done.

brownsarahm:
No worries, there are a lot of features here and we are happy to help you learn.  Thanks for your cooperation in getting this suggestion implemented!

brownsarahm (approved):

brownsarahm:
Congrats again on your first contribuion @martinosorb !!

PR #1454, opened 2022-10-27
Update 20-performance.md

### LINKED ISSUE #1452: rubric link in episode 20
A note that episode 20 of instructor training - https://carpentries.github.io/instructor-training/20-performance/index.html - links to the rubric over in an instructor training sandbox repo - https://data-lessons.github.io/instructor-training/demos_rubric/ .

I think it probably should link to the rubric in this repo - https://carpentries.github.io/instructor-training/demos_rubric/

ha0ye:
use link to rubric page from this repo, fixes #1452


brownsarahm:
@karenword  is there a reason this link is not a markdown link?

karenword:
This link should be both clickable and pasteable into Etherpad when the text is copied. I think this is the best way to accomplish both? But feel free to update if there is a better way @brownsarahm !

PR #1471, opened 2023-01-19
Closed 2023-01-25-ttt-online-EST

klbarnes20:


karenword:
@zkamvar we should discuss how to manage our events page in the context of the beta. I know there are contingencies for emergency updates (like event changes) but it is also nice to be able to routinely remove events from the list when they close.

zkamvar:
> @zkamvar we should discuss how to manage our events page in the context of the beta. I know there are contingencies for emergency updates (like event changes) but it is also nice to be able to routinely remove events from the list when they close.

This is a good point. I think this might just be something that I attend to manually, so please tag me whenever these event closers happen. I will write up a description of what I do to resolve it. The process will involve unlocking the branch, making the change, and then locking the branch again.

PR #1474, opened 2023-01-25
README: point maintainer links to their ID

zkamvar:
The reference links for the maintainers all go to the same place, which is not great for accessibility and can be kind of frustrating when you want to find a specific maintainer and you have to scroll through the same page over and over. 

Since the instructors page has anchors that represent the GitHub IDs for the maintainers, I've added those to the URLs to make them unique. 

This will also help address the transition to The Workbench: https://github.com/fishtree-attempt/instructor-training/issues/11 and https://github.com/carpentries/lesson-transition/issues/15


karenword:
thanks @zkamvar !

PR #1491, opened 2023-02-17
delete duplicate from Instructor Notes

karenword:
it worked! cleaning up from #1489

### DIFF
--- a/instructors/instructor-notes.md
+++ b/instructors/instructor-notes.md
@@ -237,18 +237,6 @@ This is a place for Trainers to leave tips and observations for those newer to t
 
 ### Building Skill with Practice
 
-People often request to see more examples of concept maps. These are some examples linked from a previous version of the curriculum:
-
-- [Array Math](fig/array-math.png)
-- [Conditionals](fig/conditionals.png)
-- [Creating and Destroying Files](fig/create-destroy.png)
-- [Sets and Dictionaries in Python](fig/dict-set.png)
-- [Input and Output](fig/io.png)
-- [Lists and Loops](fig/lists-loops.png)
-- [Git Version Control](fig/git_concept_map.png)
-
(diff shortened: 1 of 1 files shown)


ndporter:
I notice that while at least some of the inline instructor notes (e.g. Ep. 15) show up under the lesson headings in the main [Instructor Notes](https://preview.carpentries.org/instructor-training/instructor/instructor-notes.html) page, the changes from #1489 referenced here are only visible in the lesson page itself.

@zkamvar should in-lesson instructor notes be auto-deploying to the notes page, or does that have to be manual? 

If it's manual, we may need to have a conversation among maintainers about whether to propogate them by hand as part of any PRs that edit instructor notes or to remove the lesson headers in the main instructor notes page. If automatic, we need to know how to make sure that happens.

karenword:
Actually @ndporter I think this is working as intended for me, so perhaps there is an issue with the way you are viewing this page. I currently see this content in two locations on the rendered instructor notes page. This PR removes the first of these two screenshots: 
image
image

However, I also see that in the PR itself, the text that is auto-propagated is not visible or editable in the Markdown file for this page. So if that's where you're looking, it would appear that the duplicated content isn't there (when it actually is).

zkamvar:
> @zkamvar should in-lesson instructor notes be auto-deploying to the notes page, or does that have to be manual?

It auto-deploys. You can find the list of instructor notes from each episode in the `aggregate-instructor-notes` section: https://preview.carpentries.org/instructor-training/instructor/instructor-notes.html#aggregate-instructor-notes

Thank you for reminding me that I need to update the documentation for the differences and for a general overview on how the aggregate pages work. https://carpentries.github.io/workbench/transition-guide.html#instructor-notes

Additionally, you can delete the episode headings at the bottom of the instructor notes page that you do not need.

ndporter:
Thanks @karenword; I think this is ready to merge now?

PR #1492, opened 2023-02-17
inline instructor note for captions,

### LINKED ISSUE #1399: Direct trainers to have live captions on by default
## 6. Useful adaptations – things which went well

Live captioning in Zoom was incredibly useful. I only realised that I could request this on the afternoon of the second day. This allows participation with the sound turned off (addressing auditory sensitivity), and to use the transcript to catch up if a momentary loss of concentration means that content has been missed.

Perhaps trainers can be encouraged to turn this on by default! I also noticed however that when Zoom breakout rooms are used the subtitles turn themselves off and have to be manually reactivated when returning to the main room. The use of breakout rooms did really gave a sense of meeting people, however, which was great.

_Originally posted by @sjmf in https://github.com/carpentries/instructor-training/issues/1395#issuecomment-1163278874_

brownsarahm:
closes #1399

### DIFF
--- a/episodes/01-welcome.md
+++ b/episodes/01-welcome.md
@@ -23,7 +23,7 @@ exercises: 15
 
 
 :::::::::::::::::::::::::::::::::::::::instructor
-If on zoom, ensure that captions are turned on
+In an online workshop, ensure that captions are turned on.
 
 
 :::::::::::::::::::::::::::::::::::::::::::::::::::::::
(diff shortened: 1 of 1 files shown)


ndporter:
I made some minor edits to avoid platform-specific language. Along with this, we should also find a place in the episode to have instructors mention that captions are enabled and how to turn them on (maybe with an instructor note on the current Zoom process).

karenword:
I can't figure out where to add instructions in the curriculum so I created an issue for your point @ndporter and merging this for now.

PR #1494, opened 2023-02-21
Update 02-practice-learning.md

mjy:
George Box did not make the literal quote that is attributed to him here.  If it is summarized as it currently exists perhaps it should include elipses to indicate portions that are removed or altered?

### DIFF
--- a/episodes/02-practice-learning.md
+++ b/episodes/02-practice-learning.md
@@ -93,7 +93,7 @@ for how we teach.
 
 :::::::::::::::::::::::::::::::::::::  testimonial
 
-All models are wrong, but some are useful.
+Remember that all models are wrong; the practical question is how wrong do they have to be to not be useful.
 
 - George Box, statistician
   


brownsarahm:
admittedly we did not have a source previously, but a quick search shows that he did say "all models are wrong" in [this paper](http://www-sop.inria.fr/members/Ian.Jermyn/philosophy/writings/Boxonmaths.pdf)

and though I cannot access the proceedings, [wikipedia says he titled a section of a paper with that quote](https://en.wikipedia.org/wiki/All_models_are_wrong#:~:text=the%20proceedings%20of%20a%201978%20statistics%20workshop.%5B3%5D%5B4%5D%20The%20paper%20contains%20a%20section%20entitled%20%22All%20models%20are%20wrong%20but%20some%20are%20useful%22.)

If you can verify that these are untrue and provide alternative references, we can re-open the PR.

mjy:
Thanks for the leads. For the record, there is indeed a section from the cited reference (without comma, but that is pedantic):

Screenshot 2023-02-21 at 9 36 10 AM

PR #1500, opened 2023-03-01
add trainings dates for 2023 Q2

maneesha:


zkamvar:
@klbarnes20 I've put this through on the styles version as well.

karenword:
thanks @zkamvar !

PR #1521, opened 2023-04-04
take lesson out of beta

zkamvar:


zkamvar:
A little note on why this failed: it was because I was goading @karenword into merging this before the comment appeared. This is message is a failsafe for pull requests that are trying to spoof other pull requests by writing their own github pull request workflow. Because she merged this before the bot could analyze the results, it panicked and sent this message: https://github.com/carpentries/instructor-training/actions/runs/4612579850/jobs/8153636982

PR #1523, opened 2023-04-06
Removed note about schedule

arieldeardorff:
Schedule is only in the instructor view


karenword:
thanks @arieldeardorff !

PR #1529, opened 2023-04-19
Update demo etherpad link

ndporter:
Replaces no longer working `{{page.demopad}}` link with hardcoded etherpad link. Credit to Clifford Kravit for bringing this to my attention.

Need to verify with @zkamvar prior to merging whether this should be hard-linked or should remain as a page link and corrections made to ensure it works properly again.

### DIFF
--- a/learners/checkout.md
+++ b/learners/checkout.md
@@ -250,7 +250,7 @@ work for a checkout demo. Be sure to give yourself time to change course if your
 
 ### Sign-up and Set-up
 
-To sign up, select a session that works for you on [the Instructor Training Demonstration Sessions Etherpad]({{page.demopad}}), and add
+To sign up, select a session that works for you on [the Instructor Training Demonstration Sessions Etherpad](https://pad.carpentries.org/teaching-demos), and add
 your name and a link to your lesson of choice to that Etherpad. Be sure to **double check the time in your local time zone** by clicking on the converter link posted. Also, examine the demo description to ensure that it is not a special session targeting a specific sub-community or
 language (unless you are part of that target group).
 


ndporter:
Also probably for @zkamvar, I'm confused on how to maintain the files for the extras, specifically the checkout instructions. The link in the curriculum extras goes to https://carpentries.github.io/instructor-training/checkout.html (no learners or instructor). From there, you can change to instructor view and get https://carpentries.github.io/instructor-training/instructor/checkout.html. But the edit link edits `/learners/checkout.md` (whether you click from learner or instructor view).  I get that the markdown files aren't rendered in the site, but I also see that there are checkout.mds in both the `learners` and `instructors` directories on Git (and all learners md's appear to also be in instructors). If I update the md file in learners (e.g. this PR), will that propagate everywhere someone might stumble into automatically? Hopefully that's clear. Thanks!

zkamvar:
>  I get that the markdown files aren't rendered in the site, but I also see that there are checkout.mds in both the learners and instructors directories on Git (and all learners md's appear to also be in instructors). If I update the md file in learners (e.g. this PR), will that propagate everywhere someone might stumble into automatically? Hopefully that's clear. Thanks!

(a side note I am tired and my response may or may not be coherent)

In short: I kinda messed up and duplicated a lot of the files in the transition. 

There really should be no duplicated files in the repository and I'm sorry for that I felel like I really borked this repository in terms of file structure and I don't really know how better to apologiese than to ssay I'm sorry.

The best thing to do would be to look at the config.yaml and delete the offending files from these folders. I would make a PR but I currently have > 50 GitHub notifications that are screaming for my attention. 

For example, the checkout.md is mostly a file that's important for learners, so that's where it serves from. 

In terms of learner/instructor view: all files are actually visible from both views. I will be happy to explain this further when I am less messed up in the head

karenword:
thanks @zkamvar and the whole point of the beta was that we signed up for a potentially sub-optimal transition so that we could help troubleshoot. I was pretty surprised we didn't find more bug/features in the process!

It sounds like we can (and should) delete files that are available in the Learners menu from the Instructors folder, and vice versa. That is definitely something we can handle!

zkamvar:
Yes, this is the list it should be: https://github.com/carpentries/instructor-training/blob/cc587a2a0e0997654887a8f1f37cc88c2cc4693e/config.yaml#L88-L103

ndporter:
Addressed duplicates issues in #1532 . The original PR here (to fix the broken etherpad link) should still be reviewed and merged if correct.

karenword:
sorry about the duplication, and thanks for the catch!

PR #1531, opened 2023-04-20
More links to fix

karenword:
I am wondering about the link to the Etherpad that had to be replaced here, and whether we are going to find other problems related to that one. I can't recall where those links used to live in the repository. I will merge this promptly because it's important that these links work, but also tagging @carpentries/core-team-curriculum in case of advice!

### DIFF
--- a/learners/checkout.md
+++ b/learners/checkout.md
@@ -296,7 +296,7 @@ Approximately 1-2 weeks after your last step is complete, you should receive an 
 
 [help-wanted]: https://carpentries.org/help-wanted-issues/
 [github-guide]: https://github.com/carpentries-incubator/swc_github_flow/blob/master/for_novice_contributors.md
-[contributing]: https://github.com/carpentries/instructor-training/blob/gh-pages/CONTRIBUTING.md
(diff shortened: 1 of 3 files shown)


learners/checkout.md
```diff
@@ -250,7 +250,7 @@ work for a checkout demo. Be sure to give yourself time to change course if your
 
 ### Sign-up and Set-up
 
-To sign up, select a session that works for you on [the Instructor Training Demonstration Sessions Etherpad]({{page.demopad}}), and add
```
zkamvar, line 253: Sorry about that. This was one of the unique features of the IT curriculum that I did not catch.

This in particular was taking the `demopad` parameter from the YAML header and replacing it in this link. 

This was a case where a Jekyll solution was found where a more general (and portable) markdown solution existed.

This was the Jekyll solution:

```markdown
---
demopad: 'https://pad.carpentries.org/teaching-demos'
---

[Etherpad]({{ page.demopad }})
```

But it breaks in a non-Jekyll environment. Here's the general markdown solution:

```markdown
[Etherpad][demopad]

[demopad]: 'https://pad.carpentries.org/teaching-demos'
```

PR #1532, opened 2023-04-21
Delete duplicate extras

ndporter:
Removes duplicated or unnecessary files in `/learners` and `/instructors` directories. Only one version of each is necessary following the workbench transition. List to keep taken from `config.yaml` (lines 88-103) per comment in #1529.


karenword:
thanks @ndporter !

PR #1547, opened 2023-05-30
fix url closes #1546

### LINKED ISSUE #1546: broken link on " Suggested Rubric for Teaching Demonstrations" page
https://carpentries.github.io/instructor-training/demos_rubric 

on this page, the "good lessons for demo" link in the first paragraph is broken.

brownsarahm:
I think fixes, things, but also I wonder if this should have been a relative link that is rendered, instead of typed out this way.

### DIFF
--- a/learners/demos_rubric.md
+++ b/learners/demos_rubric.md
@@ -2,7 +2,7 @@
 title: Suggested Rubric for Teaching Demonstrations
 ---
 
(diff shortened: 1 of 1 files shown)


brownsarahm:
related to #1522  and #1531

karenword:
Thanks @brownsarahm -- and probably yes, but we can do a sweep to find and convert to relative links at some point if it seems like a good idea.

Issue #1363, opened 2022-05-02
Participatory Live Demonstration vs Coding


brownsarahm:
From Nathaniel Porter:
> Had *[instructor trainees who do not program in instructor training workshop this week]* and got feedback that *[most of the IT curriculum]* assumed some form of coding as the primary goal. They noticed there wasn't really any discussion of differences in live-coding when you're using a graphical tool (like the OpenRefine and Spreadsheets lessons) instead of a command-line or programming tool.

*[denotes changes by me]*

This issue is to collect notes about how trainers talk about teaching these tools in Instructor Training workshops and and tips from any instructors who teach these tools often so that they can be incorporated into the curriculum. The final format of inclusion will be determined based on the tips we get.  

For me the main difference is that **go slowly** is even more important (learners can't catch up by copy-pasting) and it looks a little different: 
- hover over menus and then explain some of what is going on or conceptually how and why things are nested as they are before clicking to give learners time to click on the menu items as well. 
- carefully describe where on the screen items are and what shape/designing features they have in addition to their actual name( eg the github "create a new..." menu is in the top right of the screen to the left of your profile picture. it's a plus sign with a down arrow next to it). this both helps people find it and slows you down giving them more time to do so. 
- use stickies for all learners to indicate that their screen looks like yours

ragamouf:
Giving public/"traditional" librarians a novel experience to teach live coding is a salient example to bring learner doubts and insecurities to the fore, while they learn explicitly about teaching skills. 

As a coding novice myself, running commands in the command line is sufficiently different to "yet another tool" that as a learner I have a richly relatable experience that I can apply to what I'm learning about in how to teach others in an area they may feel conspicuously amateur.

It's important how you frame the learning task. Perhaps it's worth acknowledging common misconceptions or challenges for the usefulness of coding for "traditional" Librarians. Such as:

1.  Why should I learn how to code and even teach our users/patrons/clients to code when we have an IT team?
2. Coding skills will be redundant in the future because we'll have tools and graphical user interfaces for that.
3. It's too hard, my brain doesn't work that way.
4. Coding is only for computer scientists.

All these challenges can be useful talking points as you introduce either participatory live coding or live demonstration.

brownsarahm:
[...]

PR #1555, opened 2023-06-08
Remove 'programming'

fiveop:
The Carpentries do not only teach programming and use participatory live coding to teach other skills as well. Thus, we should not specifically mention programming in the overview question.


ndporter:
This PR makes sense, as programming is not even mentioned in the remainder of the lesson and it is only a subset of what we teach. Ideally (per #1363 ) we would also potentially update the whole episode to minimize "coding" specific references, since there are multiple lessons in both library and data carpentry that don't involve coding, even in a broad sense (spreadsheets, OpenRefine, etc.).

PR #1558, opened 2023-06-09
Update live code to live demos in 17-live.md

ndporter:
1. Replace "coding" in most instances with "demo" or related words to make more general
2. Update line 33 to reflect both online and in-person workshops
3. Remove wikipedia "live coding" link as it refers to live coding as a research or artistic tool, but doesn't reference pedagogy at all
4. Added callout on live coding/demos


episodes/17-live.md
```diff
@@ -1,51 +1,67 @@
...
 ---
 
 One of the cornerstones of The Carpentries teaching is live
-coding: *instructors do not use slides to teach coding*, but work through the lesson material,
-typing in the code or instructions, with the workshop participants following
+demonstrations: *instructors do not use slides to teach*, but work through the lesson material,
+typing in code or modeling procedures, with the workshop participants following
 along. This section explains how it works, why we use it, and
-gives general tips for an effective participatory live coding presentation. We will
+gives general tips for an effective participatory live demo presentation. We will
```
brownsarahm, line 13:
```suggestion
gives general tips for an effective participatory live demonstration. We will
```
"demo presentation" reads funny to me

episodes/17-live.md
```diff
@@ -174,9 +190,9 @@ This exercise should take about 25 minutes.
 
 :::::::::::::::::::::::::::::::::::::::: keypoints
 
-- Live coding forces the instructor to slow down.
+- Live demos forces the instructor to slow down.
```
brownsarahm, line 193:
```suggestion
- Live demonstration forces the instructor to slow down.
```
or live demo maybe, but "live demos" feels like a mismatch of plurality?

episodes/17-live.md
```diff
@@ -107,7 +123,7 @@ Many error messages are obscure and not written with novices in mind. Continue t
 
 ## Compare and Contrast
 
-Watch this first participatory live coding demo video: [https://youtu.be/bXxBeNkKmJE][live-coding-bad]
+Watch this first participatory live demo video: [https://youtu.be/bXxBeNkKmJE][live-coding-bad]
```
brownsarahm, line 126: this is actually code, so I think here, live "coding" is actually more accurate and we should augment the exercise with reflection on what we can learn from this that applies to non-coding contexts

episodes/17-live.md
```diff
@@ -90,9 +106,9 @@ Some challenges are:
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
-Live coding fits well into the practice-feedback model we have been discussing - by providing
-learners with continuous opportunities for practice (every time they type in a line of code)
-and continuous feedback (their code either works or fails with an error message). It is
+Live demos fits well into the practice-feedback model we have been discussing - by providing
+learners with continuous opportunities for practice (every time they type a line of code or perform a procedure)
+and continuous feedback (their attempt either works or fails with an error message). It is
```
brownsarahm, line 111: the "error message" part may not translate to non-coding contexts, but i'm not sure what the best fix is here.

episodes/17-live.md
```diff
@@ -90,9 +106,9 @@ Some challenges are:
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
-Live coding fits well into the practice-feedback model we have been discussing - by providing
-learners with continuous opportunities for practice (every time they type in a line of code)
-and continuous feedback (their code either works or fails with an error message). It is
+Live demos fits well into the practice-feedback model we have been discussing - by providing
```
brownsarahm, line 109:
```suggestion
Live demonstration fits well into the practice-feedback model we have been discussing - by providing
```
i think using the full work is better for non native english speakers, but would accept the singular of the shortcut

brownsarahm (changes requested):
I picked at a few places specifically, but I actually think that we should use "demonstration" throughout because full words are more inclusive than shortenings for non native english speakers, I think. 

It's also more consistent to use the exact same term throughout instead of two forms of it.

episodes/17-live.md
JCSzamosi: Error messages defnitely come up in non-coding contexts IME.

ha0ye:
@brownsarahm - I am internally amused about people reading "live demo" and then expecting to see "Carpentries Live Demolitions", but your suggestion is excellent!

episodes/17-live.md
```diff
@@ -1,51 +1,67 @@
...
+:::::::::::::::::::::::::::::::::::::::::  callout
+
+## Live Coding or Live Demonstration?
+
+The term "live coding" is used throughout instructor training
+to describe the Carpentries model of teaching,
+but you don't need to be a programmer or know how to code at all
+to integrate live demonstrations in your teaching.
+Many of our lessons on tools like spreadsheets and OpenRefine
+include no coding at all, and similar models are widely used
```
villanueval, line 55: This is not entirely correct, although I see the distinction of these two lessons. Both use some use of functions. Might change it to "Some lessons might not include coding..."

ragamouf:
Following today's discussion in the trainer meeting, my consideration into this change comes to this:
If we teach a cohort of trainees with no experience coding, programming, developing software, then it is up to us to explain the place of live coding in our pedagogy, and our emphasis on "participatory live coding". I don't agree with the wholesale change to replace 'coding' with 'demonstration' or other general presentation forms because it takes the learner away from the new mental model that we are teaching, which is associated explicitly with programming skills. When words and concepts are unfamiliar to learners, we can teach them to embrace the discomfort.

JCSzamosi:
The reason I don't fully agree with this is that we do, in fact, have quite a bit of curriculum that does not involve coding. The open refine and spreadsheet lessons in the DC and LC curricula, at least. I would even go so far as to say that the Git lessons don't strictly speaking teach coding, but rather interactively interacting with Git on the command line.

The point, as I understand it, is not to avoid the word "coding" because it might be unfamiliar or intimidating, but to specifically be inclusive of people who are teaching those lessons, and to avoid confusion by making it clear that the live participatory stuff that we do is not limited to coding.

ndporter:
Closing after creating issue #1779 to encourage revisiting this at a later time.

Issue #1554, opened 2023-06-08
Broken links in Instructor Training - Recommended episodes


vansteph:
Hey,

the links to the recommended episodes in the instructor training preparation document (https://carpentries.github.io/instructor-training/#recommended-episodes) are all broken as they link to the old lesson design.

All the best!

Issue #1549, opened 2023-06-02
On "Summary and Setup" page, link to "Tracking Changes in git" is broken


allie-tatarian:
Under the [Recommended Episodes]( https://carpentries.zoom.us/my/carpentriesroom2?pwd=WmVCOUlPUm1laFk5SUp1UWg5cjhEUT09), "Tracking Changes in git" currently leads to a [Page Not Found](https://swcarpentry.github.io/git-novice/04-changes/) within the git lesson.

sjvrijn:
I believe this issue can be closed, since it seems like #1559 has indeed fixed the broken link(s) on the Summary and Setup page under [Recommended Episodes](https://carpentries.github.io/instructor-training/index.html#recommended-episodes)

Issue #1542, opened 2023-05-24
In the demo lessons page, the links in Astronomy section appear to be broken


cidule:
Demo lessons markdown I am referring to:
https://github.com/carpentries/instructor-training/blob/main/learners/demo_lessons.md

In the Astronomy section (https://carpentries.github.io/instructor-training/demo_lessons#dc-astro) the links in the "Good starting points" and "Notes" columns appear to be broken.

PR #1559, opened 2023-06-10
Fixed faulty carpentry workshop URLS

utsmok:
Because of the new format of the workshops, some links were dead. I fixed all links that pointed towards carpentry workshops, afaik.  Fixes issues #1554 #1549 #1542.


karenword:
thanks @utsmok !

PR #1570, opened 2023-07-17
fix typos

mhennies:
typo in quotes


ndporter:
Good catch on the quotes. I added an update to capitalization for Bash, so I'll wait for someone else to review before merging.

PR #1581, opened 2023-08-14
Remove stray “Welcome session”

fiveop:
As the title says, the patch removes the words Welcome session that seem lost between two sentences.


karenword:
Thank you @fiveop !!

PR #1583, opened 2023-08-14
Make framework recognize list

fiveop:
This adds an empty line between a paragraph and the following list. Previously, the list content was rendered as part of the paragraph instead of as a list.


klbarnes20:
Thanks for all the fixes @fiveop!

PR #1591, opened 2023-09-08
Remove registration_confirmation.md from menu

### LINKED ISSUE #1486: Remove Registration Confirmation Page from menus
[this page](https://preview.carpentries.org/instructor-training/instructor/registration_confirmation.html) doesn't need to be in the Instructors OR the learners' menu -- it is a page that exists only to be linked from Eventbrite. 

Will it break anything if we just take it off the list in the yml? @zkamvar

ndporter:
Closes #1486 by moving registration confirmation page from `instructors` to root. Also removes from config.yaml.


ndporter:
Can someone (maybe @zkamvar ?) verify that nothing else depends on this file being located where it was previously (in `instructors`) before merging?

zkamvar:
> Can someone (maybe @zkamvar ?) verify that nothing else depends on this file being located where it was previously (in `instructors`) before merging?

If any other file depended on this, you would get a new warning in [the preview build](https://github.com/carpentries/instructor-training/actions/runs/6123576707?pr=1591), which would also show up in [the files preview](https://github.com/carpentries/instructor-training/pull/1591/files)

Just to be sure, I've checked this manually:

```sh
# in the shell
$ gh pr checkout 1591
```

```r
library("pegboard")
lsn <- Lesson$new(jekyll = FALSE)
lsn$validate_links()
```

No messages; no problems. 

> I also notice the `permalink` field in the header with `/registration/` that I'm not sure we need anymore.

No, that is not really needed anymore. That was used by Jekyll, but not by The Workbench

ndporter:
Thanks @zkamvar for checking and for sharing the code on how you checked. I've pushed another commit that removes the `permalink` line and I think this is ready to go.

brownsarahm (approved):

PR #1597, opened 2023-09-20
Fix text in headers

karenword:
I noticed that the Workbench had positioned some content from the top of each page *above* the lesson questions and objectives. In a couple of cases this kinda worked, but in most cases it really didn't. I changed them all -- in the future we can consider whether we want to exercise the option of positioning a summary of some kind above the questions/objectives box.

@zkamvar FYI in case this affects other lessons - sorry I didn't notice this during the beta!

### DIFF
--- a/episodes/02-practice-learning.md
+++ b/episodes/02-practice-learning.md
@@ -5,10 +5,6 @@ teaching: 30
 exercises: 30
 ---
 
-We will now get started with a discussion of how learning works. We will begin with
-some key concepts from educational research and identify how these principles
-are put into practice in Carpentries workshops.
-
 ::::::::::::::::::::::::::::::::::::::: objectives
 
 - Compare and contrast the three stages of skill acquisition.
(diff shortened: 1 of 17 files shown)


zkamvar:
Thank you for tagging me, Karen. This is only specific to the four beta phase repositories. I had fixed this during the transition after one of the eagle-eyed maintainers (I believe it was a LC maintainer) pointed it out

News item announcing the fix: https://carpentries.github.io/pegboard/news/index.html#pegboard-053-2023-07-08
Relevant PR in {pegboard}: https://github.com/carpentries/pegboard/pull/124

zkamvar:
And sorry for the kerfuffle and confusion about it

dpshelio (approved):
It's not clear to me whether we should merge this or the changes in `{pegboard}` will propagate to this too.

zkamvar:
> It's not clear to me whether we should merge this or the changes in `{pegboard}` will propagate to this too.

This should be merged; it's related to the transition, not how {pegboard} works day to day.

PR #1599, opened 2023-09-29
remove "workshop"

karenword:
Cleaning up a couple of instances of "workshop" to describe the event, as we are supposed to use "training" for distinction from technical workshops.


dpshelio (approved):

dpshelio:
We can merge #1606 first to get all these changes together.

PR #1604, opened 2023-10-17
replace Twitter with Mastodon

elichad:


episodes/15-carpentries.md
```diff
@@ -279,7 +279,7 @@ ways to keep in touch, and we do our best to make that possible.
 Want to **listen**?
 
 - Sign up for our newsletter
-- Follow us on Twitter, Facebook, or LinkedIn
+- Follow us on Mastodon, Facebook, or LinkedIn
```
acrall, line 282:
```suggestion
- Follow us on Mastodon or LinkedIn
```

acrall (changes requested):
We are moving away from using Facebook so I suggested that Facebook be removed as well

acrall (approved):
Perfect, thanks for the help in getting this updated!

PR #1605, opened 2023-10-20
correct callout type

karenword:
This quote is not a testimonial. 

Fixing this one only, but note that there may be other errors of this kind, because the prior template did not label the callouts and were sometimes selected based on color rather than strict categorical matching!

### DIFF
--- a/episodes/02-practice-learning.md
+++ b/episodes/02-practice-learning.md
@@ -91,7 +91,7 @@ for how we teach.
 
 ## Building a Mental Model
 
-:::::::::::::::::::::::::::::::::::::  testimonial
+:::::::::::::::::::::::::::::::::::::  callout
 
 All models are wrong, but some are useful.
 


dpshelio (approved):
I think that's the only one, the other two seems real.

```bash
instructor-training/episodes on karenword-patch-2 
❯ rg -A 5 testimonial
02-practice-learning.md
434::::::::::::::::::::::::::::::::::::::  testimonial
435-
436-If someone feels it is too slow, they will be a bit bored. If they feel it is too fast, they will never come back to programming.
437-— Kunal Marwaha, SWC Instructor
438-
439-

08-motivation.md
184::::::::::::::::::::::::::::::::::::::  testimonial
185-
186-## Typos
187-
188-The typos are the pedagogy.
189-— Emily Jane McTavish
```

ndporter:
I'm going to merge this as-is, but I made an issue (#1608 ) to check the type/structure of various special sections and ensure it's consistent. I think this is probably low-priority, but it's there as a reminder.

PR #1606, opened 2023-10-22
workshop to training

dpshelio:
This continues  #1599 but for all the files.

@karenword - if this is merged is added to your previous PR. I wanted to make sure I did it right.
Note that my second commit, instead of replace `workshops` I added `activities`  as it refers to where the CoC takes effect.


dpshelio:
Also, not sure whether we want to change it here too:  https://github.com/carpentries/instructor-training/blob/7a642a0e4fada65f3e2f2b0e0fc9839ab9f38bce/episodes/25-wrap-up.md#L52-L56

PR #1613, opened 2023-11-03
Update 09-eia.md

ndporter:
Add an instructor note to accessibility section highlighting that accessibility can affect entire groups or regions, not just (disabled) individuals.

I'm not sure if this is the best way to word it or the best place to set this note in the lesson (or even if it should be in the main text instead of an instructor note). This PR came about because of responses by instructor trainees to [this activity](https://carpentries.github.io/instructor-training/instructor/09-eia.html#what-happens-when-accessibility-is-an-issue).

### DIFF
--- a/episodes/09-eia.md
+++ b/episodes/09-eia.md
@@ -69,6 +69,15 @@ This exercise should take about 5 minutes.
 Barriers to accessibility encountered in a workshop are demotivating at best and can exclude a learner entirely.
 What barriers might be present in your workshop, and what can you do to remove them?
 
+::::::::::::::::::::::::::::::::::::::: instructor
+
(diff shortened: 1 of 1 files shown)


episodes/09-eia.md
```diff
@@ -69,6 +69,15 @@ This exercise should take about 5 minutes.
 Barriers to accessibility encountered in a workshop are demotivating at best and can exclude a learner entirely.
 What barriers might be present in your workshop, and what can you do to remove them?
 
+::::::::::::::::::::::::::::::::::::::: instructor
+
+In your discussion of accessibility, it is helpful to point out that accessibility may not alway affect only individual learners,
```
karenword, line 74:
```suggestion
In your discussion of accessibility, it is helpful to point out that accessibility may not always affect only individual learners,
```

PR #1614, opened 2023-11-03
Clarify wording on recovering motivation

ndporter:
A suggestion to replace confusing (to me) "rescue" metaphor in sentence on managing motivation with more direct language.

This one may just be me but it takes me a minute to process the sentence 

> What is important is to be mindful of times when your content is not motivating, and to strategize your rescue using some of the other techniques in this section.

because the word rescue throws me off, plus I don't think it always needs something as dramatic as rescue. I suggested some new language that is more direct and to me, less confusing.


karenword (approved):
lgtm -- thanks!

PR #1616, opened 2023-11-03
Point to all checklists and make helper recommendations consistent

ndporter:
1. Update callout on local help in Carpentries episode to point to the full collection of workshop checklists for local teams (not just helper checklist).
2. Change recommended ratios to be consistent with handbook. If the handbook is incorrect and the original numbers are consistent with current recommendations, this can be returned to 1:8 (in-person) and 1:5 (online) but the handbook checklists should then be updated.

### DIFF
--- a/episodes/15-carpentries.md
+++ b/episodes/15-carpentries.md
@@ -114,8 +114,8 @@ certified instructors.
 
 ## Local Support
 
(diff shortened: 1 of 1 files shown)


karenword:
@sheraaronhurt can you confirm that these are the most current recommendations for helper-to-learner ratios?

sheraaronhurt:
@karenword  yes that is correct. Thanks for checking!

PR #1617, opened 2023-11-03
Update list of community roles to match handbook

ndporter:
Updates the list of roles in community roles activity to match the names, roles and capitalization currently included in the page learners are instructed to refer to (https://carpentries.org/community/).


karenword:
Tagging @acrall because I'm not actually sure the page we are point at is up to date either. Any suggestions for addressing the flux of available roles?

ndporter:
@karenword if we want to future-proof this, the easiest thing might be to give a shorter list of roles that are unlikely to be removed (and ideally unlikely to change names). 

The question now asks for:

> 1) a short definition of the role and 
> 2) a question that you have (or that you imagine someone else might have) about the role. Are there roles you would like to see that are not listed? Note that, too!

It would be easy enough to revise part 2 a bit to something like:

> 2) a question that you have (or that you imagine someone else might have) about the role. Are there other Carpentries roles on the webpage (or roles you think might be helpful) that are not listed? Note that, too!

The only real downside to that is that the more specialized roles might not get noticed as much if trainers don't make a point to call attention to them.

acrall:
@karenword Most of this content will be changing with the rollout of our new website. We did not want to spend a lot of time on updates to the existing site as that process was happening (although there have been MANY delays that have upended this approach). We have been developing a workflow for the new website that assigns pages to our various teams to review and update the content periodically so things should not get so outdated moving forward. If we want to go ahead and transition the content on this page to what will appear on the new website, I can do that after discussing with the website working group. Our next meeting is on Friday

episodes/15-carpentries.md
```diff
@@ -259,13 +259,13 @@ Select one role from the list below that interests you. Using the the descriptio
...
-6. Instructor Development Committee
-7. Community Facilitators
-8. Maintainers
+2. Instructor development
+3. Trainers
+4. Lesson developers
+5. Curriculum advisors
+6. Lesson maintainers
+7. Regional Coordinators
+8. Code of Conduct Committee
```
karenword, line 268:
```suggestion
2. Instructor Trainers
3. Lesson developers
4. Curriculum advisors
5. Lesson maintainers
6. Code of Conduct Committee
```

karenword: I think this list may be more accurate. Instructor development has been retired, and while there are still people who are serving as regional coordinators I do not think this is continuing as a formal role.

PR #1618, opened 2023-11-03
Add instructor note on learner introductions

ndporter:
Add instructor note discussing more options for inviting learners to get to know one another, as well as highlighting needs to potentially hear names pronounced and to balance time needs.


karenword:
thanks @ndporter !

PR #1626, opened 2023-11-22
Clearify exercise about praises

cknuepfer:
The example praise 5 is based on ability. But "ability" was not mentioned in the question. On of our participants find this confusing.


brownsarahm:
The solution states that 5 is an example of performance-based praise.  This change would make the exercise not match the solution. 

Above, in the text, the lesson states "performance or ability" with the intention that "ability" is included in performance based praise.  


I'm going to close this, but a different type of clarification is welcome.

PR #1628, opened 2023-11-24
Replace teach/teaching by introduce/introduction

fiveop:
In this exercise, participants are giving introductions to an episodes, which is part of teaching, but to avoid confusion I suggest to consistently refer to the activity by introduce/introduction.


karenword:
thanks @fiveop !

PR #1629, opened 2023-11-24
Suggest to sign-post Q&A periods

fiveop:
We could suggest to sign-post Q&A periods, because it can help learners to mentally prepare for them and not feel put on the spot.

This should, however, not suggest that you should not allow questions at any time, which for some might be implied by the suggested text.

This was a suggestion by a participant in our recent instructor training.

### DIFF
--- a/episodes/04-expertise.md
+++ b/episodes/04-expertise.md
@@ -247,6 +247,7 @@ moving on, the more this implication is magnified.
 Instead, consider asking "What questions do you have?" and leaving a healthy pause for consideration.
 This firmly establishes an expectation that
 people will, indeed, have questions, and should challenge themselves to formulate them.
+By additionally sign-posting upcoming Q&A periods, learners can mentally prepare and are less put on the spot.
 
 ## You Are Not Your Learners
 


ndporter:
Thanks for the contribution @fiveop!

I like the idea of suggesting this, although we probably need to refine the language a little bit to make sure it's clear and understandable to everyone. I'm not sure "sign-posting" is understood across all the communities our instructors would come from.

Also, I'm wondering if this would be better as a callout (visible to everyone but set aside from main text) or trainer note (only visible in the instructor view)? Those are both commonly used for suggestions of things that trainers may want to bring up (or trainees stumble on) that may be helpful but aren't core to the lesson.

fiveop:
I agree with your first paragraph. I had to think a bit myself what was meant, when the trainee mentioned it. Another suggestion:

“By pointing out upcoming Q&A periods, we can give learners the opportunity to mentally prepare and be less put on the spot.”

Regarding the second paragraph. I think that putting it in a callout or trainer note would give it more visibility than it deserves compared to the other things mentioned in the paragraph. This way it is just a single sentence addition, which should be as small as possible anyway, given that we usually cannot cover all material in our trainings.

episodes/04-expertise.md
```diff
@@ -247,6 +247,7 @@ moving on, the more this implication is magnified.
 Instead, consider asking "What questions do you have?" and leaving a healthy pause for consideration.
 This firmly establishes an expectation that
 people will, indeed, have questions, and should challenge themselves to formulate them.
+By additionally sign-posting upcoming Q&A periods, learners can mentally prepare and are less put on the spot.
```
karenword, line 250:
```suggestion
If you plan to ask for questions at specific points, you can further encourage learners to prepare questions by including checkpoints in a posted schedule or announcing them in advance. 
```

karenword: I wonder if this is a more complete explanation of the suggested idea? I do worry a bit that people will interpret this as a suggestion to *restrict* questions to Q&A periods, but I'm not sure how to avoid that without making this unnecessarily verbose.

fiveop: I am not even sure that it is enough. If you announce that there will be a time for questions, you have to explicitly state that questions can be asked at all times as well (if that is what you want), otherwise I would feel the opposite is implied.

karenword:
I think we are agreed that, while this proposed addition is a useful idea, it also introduces new problems that may demand further text to clarify. In the interests of preserving focus, I'm going to close this.

PR #1632, opened 2023-12-14
Add callout with demos video

karenword:
This is a start at incorporating the new video!


ndporter (approved):

ndporter:
I like the approach to integrating it here. We may at some point also want to add links in the checkout page and teaching demos page or etherpad. But that shouldn't stop this one going through.

PR #1643, opened 2024-02-22
Fix typo in cutting call out 

### LINKED ISSUE #1642: Typo in "A Note on Cutting"
### What is the problem?

The note starts with "This episode is a common place to Trainers to cut while preparing to teach." It would probably make more sense if it was phrased more along the lines of "This is a common episode that gets cut when preparing to teach."

Also, this note might be better placed after the page introduces "cutting" later on, rather than before. 

### Location of problem (optional)

https://carpentries.github.io/instructor-training/18-preparation.html

brownsarahm:
attempts to fix #1642 

I did something slightly different than the suggestion, but i agree there was a typo in the box. 

I think leaving it at the top actually makes sense for two reasons:
1. in theory, trainees are coming back to this page as a reference after the workshop more often than reading it when they would not know what we mean by cutting, and having a warning at the top of the episode that the version they saw in their workshop may have been truncated
2.  it serves as an additional reminder to trainers that this episode is a good place to make up time if needed. While a trainer would see the entirety of the episode while preparing, I think at the top it is more likely to catch & support in real time a new trainer who might be nervous; that is the note to cut from this episode and reassurance that others cut here too, could help prevent someone from speeding up instead of cutting, but for that to work, the reminder has to be before the content that would be cut. 

If others 1) agree with that rationale 2) think it is worth making more findable,  we could put a (shortend; edited for clarity) version of that justification in the box too.

### DIFF
--- a/episodes/18-preparation.md
+++ b/episodes/18-preparation.md
@@ -41,7 +41,7 @@ we therefore suggest setting aside time *before* deeply reviewing your technical
 
 ## A note on cutting
 
-This episode is a common place to Trainers to cut while preparing to teach.
+This episode is a common place for Trainers to cut while preparing to teach.
 That's not because this is not important -- this page is a valuable resource -- but we feel this
 is one of the sections that trainees can use effectively as a resource when actually preparing
 for a workshop, even without spending a lot of time doing activities on this material during
--- a/episodes/18-preparation.md
+++ b/episodes/18-preparation.md
@@ -49,6 +49,14 @@ their Instructor Training event.
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
+:::::::::::::::::::::::::::::::::::::::::  callout
+
+## Using this page as a resource
+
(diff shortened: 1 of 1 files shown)


ndporter (commented):
Agree this is an improvement and addresses the most important parts of the issue. My only question is whether this should be an instructor note (e.g. trainer note) instead of a callout? That is, should everyone who goes to the lesson see it or only if they're using instructor mode?

brownsarahm:
That is a good point, actually with instructor view, maybe we do audience specific versions approximately like: 
- for trainers: this is a place you may want to cut ... 
- for trainees: this episode may have more detail in it that was covered in your works ...

ndporter:
@brownsarahm what do you think of the changes I proposed in response to your suggestion?

brownsarahm:
I like it, thanks! @ndporter

jonathanwheeler01 (commented):
Looks good. I will merge.

PR #1660, opened 2024-04-12
Changed to HTTPs instead of HTTP

annajiat:
tiny change of adding s

PR #1661, opened 2024-04-15
update license to match sandpaper template

brownsarahm:
After #1660 I noticed that it had copyright as SWC. but the current template says Carpenties.


tobyhodges (approved):
Good catch, thank you @brownsarahm

PR #1665, opened 2024-05-10
Add alternative to 2-day workshops

ndporter:
Adjust statement on commonality of lesson programs to indicate all share a "two to four day" format instead of just 2-day. This reflects current practice.

### DIFF
--- a/episodes/15-carpentries.md
+++ b/episodes/15-carpentries.md
@@ -86,7 +86,7 @@ Their aligned missions are accomplished by running accessible, inclusive trainin
 Similarities between Data Carpentry, Library Carpentry, and Software Carpentry workshops include:
 
 - a focus on technical skills,
-- a two-day format taught by volunteer instructors, and
+- a two to four day format taught by volunteer instructors, and
 - a focus on filling gaps in current training for learners.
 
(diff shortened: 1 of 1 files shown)


episodes/15-carpentries.md
```diff
@@ -86,7 +86,7 @@ Their aligned missions are accomplished by running accessible, inclusive trainin
 Similarities between Data Carpentry, Library Carpentry, and Software Carpentry workshops include:
 
 - a focus on technical skills,
-- a two-day format taught by volunteer instructors, and
+- a two to four day format taught by volunteer instructors, and
```
brownsarahm, line 89:
```suggestion
- a two full day or four half day format taught by volunteer instructors, and
```
a bit longer, but is this more accurate? or are some four full days?

brownsarahm: @ndporter thoughts on this?

ndporter: @brownsarahm this is probably fine. I think there are occasionally 3-day versions that sites stand up, but they're non-standard and your text is more clear on the scope (that it's not 4 full days but always the same volume of content). Thanks!

PR #1675, opened 2024-06-19
add solution linking to glossary

### LINKED ISSUE #1655: Jargon definitions needed
### How could the content be improved?

In episode 15 there is a discussion exercise about definitions of some jargon terms. There aren't definitions included in the manual at this point, and I've not been able to find any good source of definitions for those six terms. Most of them I've been able to track down somewhere, but I've yet to find a clear and concise description of perhaps the most critical, which is what a lesson is (other than a collection of episodes, but that's just a circular definition since an episode is defined as part of a lesson). Would it be possible to provide definitions, links to definitions, or consider dropping this actvity?

### Which part of the content does your suggestion apply to?

https://carpentries.github.io/instructor-training/15-carpentries.html#test-yourself

tobyhodges:
_If this pull request addresses an open issue on the repository, please add 'Closes #NN' below, where NN is the issue number._

Closes #1655

_Please briefly summarise the changes made in the pull request, and the reason(s) for making these changes._

This adds a solution box to [the _Test yourself!_ exercise](https://carpentries.github.io/instructor-training/15-carpentries.html#test-yourself), containing links to the terms as defined in [the Community Glossary](https://github.com/carpentries/community-development/blob/main/glossary.md). (HT to @kariljordan for suggesting this approach.)

Please let me know if you would prefer the terms to be defined in the solution box itself: happy to update. My rationale for linking out was to avoid content rot as definitions diverge in the glossary.

### DIFF
--- a/episodes/15-carpentries.md
+++ b/episodes/15-carpentries.md
@@ -159,6 +159,20 @@ As a class or in groups, see how many of the following terms you can define.
 
 This should take about 5 minutes.
 
+::::::::::::::::::::::: solution
+
+### Links to definitions
+These terms are defined in our Community Glossary.
+
+- [Lesson](https://github.com/carpentries/community-development/blob/main/glossary.md#lesson)
+- [Episode](https://github.com/carpentries/community-development/blob/main/glossary.md#episode-lesson-episode)
+- [Workshop](https://github.com/carpentries/community-development/blob/main/glossary.md#workshop)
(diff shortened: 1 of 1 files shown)


brownsarahm (approved):
I like linking out because it also introduces a valuable resource

kariljordan:
> I like linking out because it also introduces a valuable resource

agreed!

PR #1696, opened 2024-07-12
update episode list for lc-SQL

brownsarahm:
add intro to sql as not allowed in lc-sql and explain why all ineligible episodes are ineligible. 

I  had a trainee sign up for this episode and it was neither ineligible nor recommended, but I think it does not have enough live demonstration (really any) to be a good demo episode.


ndporter (approved):

ndporter:
Looked good when I went through them. Will leave open in case someone else wants to double check before merging.

tobyhodges:
Thanks for reporting this @brownsarahm. I think when we compiled the list of starting points, I thought the [_Introduction to DB Browser for SQLite_](https://librarycarpentry.org/lc-sql/01-introduction.html#introduction-to-db-browser-for-sqlite) section could be enough to fill five minutes. But on reflection I agree that it is better to list the episode as ineligible: even if there is five mins worth of teachable content in there, it risks being buried among all the other content that cannot be taught with live demo.

tobyhodges (approved):

PR #1707, opened 2024-07-23
typo

maneesha:


episodes/06-feedback.md
```diff
@@ -58,7 +58,7 @@ a dashboard with the results! Take care not to share this link with your learner
 
 The survey links above are only for you to preview the survey as part of
 Instructor Training. When you are teaching a workshop, make sure to share the
-links that gets generated on your workshop website. Doing so will ensure that
+link that gets generated on your workshop website. Doing so will ensure that
```
ndporter, line 61:
```suggestion
links that get generated on your workshop website. Doing so will ensure that
```

There are 2 links generated (pre and post) on the website, but this is a good catch that the words still didn't agree on singular/plural.

ndporter (approved):
With the one change I added on `06-feedback.md` these look great. Thanks for being so proactive, @maneesha !

maneesha:
Thanks @ndporter I was thinking of the survey results link, where we have just one link to view both pre/post results, but you are right. For the survey itself, there are two different links.

PR #1712, opened 2024-07-24
remove dismissive language

maneesha:
The curriculum offers some resources about using pronouns.  The language "For those unfamiliar with..." reminds some users that there is a concept they are unfamiliar with, and does not seem to have any other purpose.  

I am recommending removing that language and only offering the resources.  If we are offering the resources, we should offer them as they are, without contextualizing that some people don't know about them.

### DIFF
--- a/episodes/01-welcome.md
+++ b/episodes/01-welcome.md
@@ -45,12 +45,18 @@ One reason to avoid pressuring people to share them is to allow people to share 
 gender identity only when they feel ready to.
 It is, however, necessary for all participants to use pronouns and names as listed when participants provide them.
 
-These resources may be helpful for further reading to learn more about using pronouns:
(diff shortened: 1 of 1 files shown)


ndporter:
@maneesha I love this feedback, and thought that it could also be useful to take advantage of the `spoiler` tag to put the additional resources in a foldout and reduce scrolling. What do you think of the language and structure changes I made?

maneesha:
Thanks @ndporter . Your changes look good.

(And it's one of the forms of dismissive language that gets to me more than any other.  Beginning an explanation with _"For those of you who may not know, (insert explanation here)"_ does nothing other than to remind some people of what they may not know.  If you're going to offer an explanation, just offer it without prefacing it like that.)

PR #1726, opened 2024-09-03
add 2024 Q4 events

maneesha:


mistycaseyward12:
Hope 🤞

maneesha:
I think it's ok to merge this despite the failing check... I'll revert if not.

brownsarahm:
I'm not *super* well versed in the artifact version of using gh pages, but what I do understand I think that this error may mean that the rendered site might not reflect these changes? It's not clear to me why that step is running on a PR anyway. 

but I agree it should be okay to merge this; that helps us collect more information and then we should be able to file an issue on the repo with the script code because nothing in this PR should be causing this problem. (unless the literally astronomically low probability event of a hash collision is the cause)

maneesha:
Thanks @brownsarahm - yes, nothing broke when I merged it but the changes don't render.  
It is an issue with the artifact workflow.  We are seeing this in [other lessons](https://github.com/datacarpentry/r-socialsci/pull/521) as well.  I will take another look tomorrow.

froggleston:
Fixed in [sandpaper 0.16.7](https://github.com/carpentries/sandpaper/releases/tag/0.16.7)

PR #1732, opened 2024-09-18
Add a general information page about the training

tobyhodges:
(Tagging @carpentries/core-team-workshop-admin for review and discussion. Input also welcome from Trainers/others ❤️)

As discussed in previous Trainers meetings and internally with the Workshops and Instruction Team, this would replace the individual event websites that currently need to be built from https://github.com/carpentries/training-template.

You can preview what these changes look like at https://tobyhodges.github.io/solid-spoon/training_info.html

Some notes, relative to the current content in the training website template:

* Removes link to etherpad
* Assumes trainees will know who their Trainers will be, I guess from the initial info email they receive about the event (the one that links them to this page)?
* Removes the (commented out) suggested schedule table for a 4 half-day training. Maybe this should be added to the Instructor Notes for the curriculum?
* Presents info largely on the assumption that training will take place online via Zoom. I did include some notes about what would be different if the event takes place in-person, though.
* The page says, of the pre-training survey, that "You will receive a custom link for your event when you receive your connection information." Is this accurate?

PR #1733, opened 2024-09-18
explain how parts map onto days

brownsarahm:
I realized that this is not clear always based on previous issues and I also noticed that the schedule does not indicate days clearly anymore, instead going 16 hours.  If that is a bug of the tooling we can rephrase. 

We removed references to lunch and coffee break in the break text themselves, but if people are using the schedule for planning and we increase reliance on this schedule via #1732 I think it is worth having this written at least somewhere.  

It might be nice to actually provide this outside of instructor view, but the schedule is only for instructors so I put it there too.


index.md
```diff
@@ -54,6 +54,21 @@ Feedback on these materials is welcome as an [issue][issues] on the GitHub repos
...
+### Scheduling Notes
+
+The schedule below aggregates time over nearly 16 straight hours.  
+The curriculum is broken in to four parts, labeled with start and end episodes
+and one break episode in each part. In a 2 day training, parts correspond to 
+half days, with the "end part 1" and "end part 3" episodes used as lunch breaks; 
+in a 4 day training, parts correspond to days and the time allocated to those
+two episodes is skipped. 
+
+A common modification is to switch from one break rouhgly at the middle of each
```
tobyhodges, line 67:
```suggestion
A common modification is to switch from one break roughly at the middle of each 
```

brownsarahm:
@ndporter what do you think about adding similar text to this PR on the [training calendar](https://carpentries.github.io/instructor-training/training_calendar.html) page?

PR #1750, opened 2024-11-16
Small typo fixes and some minor rewordings

anenadic:
Fixes small typos and replaces the J word term for readability.

### DIFF
--- a/episodes/08-motivation.md
+++ b/episodes/08-motivation.md
@@ -228,7 +228,7 @@ This exercise should take about 5 minutes.
 
 ::::::::::::::::::::::::::::::::::::::::::::::::::
 
-- **Leveraging the Power of "Yet"**. A request for help might start with "I can't \_\_\_ " or "I don't understand \_\_\_ ". Depending on the attitude of the learner,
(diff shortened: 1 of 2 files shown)


brownsarahm:
The case on "the Power of Yet" is a reasonable fix. 

The second change, however, I disagree with.  The point of "the J word" is to try to reinforce that we, as trainers, should not be saying the word "just" during the training, so this is a way of trying to avoid saying it (but the parenthetical is there to help it make sense in case someone forgot).  

I will draw some more attention to this.

anenadic:
> The case on "the Power of Yet" is a reasonable fix.
> 
> The second change, however, I disagree with. The point of "the J word" is to try to reinforce that we, as trainers, should not be saying the word "just" during the training, so this is a way of trying to avoid saying it (but the parenthetical is there to help it make sense in case someone forgot).

Maybe we can put the "J word" word in quotes (it currently is not) so that it is easier to parse (as a sintagma) for people reading the material? Otherwise it is a bit unclear what the J word mean. We also use the word "just" to explain what the J word means (it is also spelled out in the material) so we cannot avoid saying it in a way.

> 
> I will draw some more attention to this.

I am easy on this one - it is a small point - whatever you think is best.

PR #1751, opened 2024-11-16
Replaces the title "callout" with an actual title

anenadic:


episodes/02-practice-learning.md
```diff
@@ -93,6 +93,8 @@ for how we teach.
 
 :::::::::::::::::::::::::::::::::::::  callout
 
+## Models are not perfect but are still helpful
```
brownsarahm, line 96:
```suggestion
### Models are not perfect but are still helpful
```

I think this is better structurally for eg screen readers, but waiting on confirmation of style

brownsarahm:
Thanks @anenadic for always doing such a thorough read before you teach!! these PRs are so valuable to the community ❤️ 

I had to clarify a bit of my understanding of the workbench before applying a tiny fix on this, sorry for the delay

Issue #1755, opened 2024-11-19
Link to Annual Report


maneesha:
We are updating links that may have changed with our new website rollout.  
That aside, the link to the annual report goes to the 2022 report.  This should actually go to the 2023 report (and soon to the 2024 report).

These reports can all be found here:  https://carpentries.org/about-us/impact/

We may want to consider just linking to that page directly instead of having to update the link in this curriculum every year, and ask trainees to read the most recent year's report.

maneesha:
Edit to add this is linked to from here: https://carpentries.github.io/instructor-training/#preparing-for-your-training

PR #1766, opened 2024-12-09
Fix report reference

julia-sprenger:
Fix broken link in learner/setup.md

### DIFF
--- a/learners/setup.md
+++ b/learners/setup.md
@@ -21,7 +21,7 @@ training_site: https://carpentries.github.io/instructor-training
 3. Please read the following before the workshop begins:
 
 - ["The Science of Learning"](./files/papers/science-of-learning-2015.pdf)
-- ["The Carpentries Annual Report"](https://carpentries.org/files/reports/AnnualReport2022.pdf)
+- ["The Carpentries Annual Report"](https://carpentries.org/reports/annual-reports/2023AnnualReport.pdf)
 
 **For Online Events:**  
 4\. Please ensure that your computer and internet are sufficient to connect via Zoom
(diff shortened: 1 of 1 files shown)


learners/setup.md
```diff
@@ -21,7 +21,7 @@ training_site: https://carpentries.github.io/instructor-training
 3. Please read the following before the workshop begins:
 
 - ["The Science of Learning"](./files/papers/science-of-learning-2015.pdf)
-- ["The Carpentries Annual Report"](https://carpentries.org/files/reports/AnnualReport2022.pdf)
+- ["The Carpentries Annual Report"](https://carpentries.org/reports/annual-reports/2023AnnualReport.pdf)
```
brownsarahm, line 24:
```suggestion
- "The Carpentries Annual Report" from [The Carpentries Impact page](https://carpentries.org/about-us/impact/)
```

per #1755 recommendation to link the page instead of having to update annually.


Maybe we should add "most recent" in somewhere to make it more clear?

brownsarahm:
@maneesha and @carpentries/instructor-training-maintainers we should merge this quickly because it's a broken link, but I wanted to check if this suggestion per #1755 is well-phrased

maneesha:
As I noted in #1755, yes, we should refer to the "most recent" annual report.  That way we do not have to update it every year, and it's clear to the trainee that they only have to read that one, not everything.

brownsarahm:
thanks @julia-sprenger for improving the PR and submitting it! I'm going to merge this now, but not close the related issue because it points to an additional location we should fix.

PR #1807, opened 2025-04-25
Add CoC reference to introductions challenge

ndporter:
Add explicit reference to Code of Conduct in solution list for what should be in an introduction challenge.


ndporter:
@brownsarahm I realized after submitting this PR that it does call attention to the CoC later in the lesson. I still feel like this might be helpful, but I'd be curious your judgment.

brownsarahm (approved):
I like this, it's short and sweet, but highlights a very important point earlier

PR #1818, opened 2025-05-22
use lesson program github.io links

maneesha:
Following our new infrastructure, all lesson program links should use the `lessonprogram.github.io` url


brownsarahm (commented):
Thanks @maneesha 

I did a full search of the repo for the old URLs and replaced in all places that referred to lessons ( I think!) so someone else should review & merge

kcranston (approved):
Thanks for doing this, @brownsarahm ! LGTM - there were a few places where the markdown wasn't rendering properly in the diff view, but when in the single file viewer, it all looks good.

PR #1845, opened 2025-05-22
British spelling

maneesha:
Updates -ize, -ization, and -or words to -ise, -isation, -our to conform to British spelling as per The Carpentries' style guide.

Note this does not change spellings in the [demo lessons page](https://carpentries.github.io/instructor-training/demo_lessons.html) because corresponding links use American spelling.  For example, [Data Visualization with ggplot2](https://datacarpentry.org/R-ecology-lesson/visualizing-ggplot.html) uses American spelling in the lesson title **and** the generated url, so changing the spelling in the lesson title may cause confusion in the generated url. That is a bigger question we will need to consider.


kcranston (approved):
Thanks, @maneesha ! (this was obviously a fair bit of tedious work - we appreciate it).

Issue #1797, opened 2025-03-19
Harmonize EIA definitions with Carpentries community glossary


ndporter:
### How could the content be improved?

The Equity, Accessibiilty and Inclusion episode [defines](https://carpentries.github.io/instructor-training/09-eia.html#definitions) EIA based on the U. Pitt. DEI glossary. However, as noted in the [2025-03-19 statement/blog](https://carpentries.org/blog/2025/03/the-carpentries-reinforced-commitment-to-inclusion-diversity-equity-and-accessibility/), our community glossary has our own definitions. 

There is a case to be made that some of the current definitions are more tightly coupled to the information in the episode and shouldn't change (or should change but with some modification), so I'm not proposing this as a PR until there's been some time for discussion.

### Which part of the content does your suggestion apply to?

https://carpentries.github.io/instructor-training/09-eia.html#definitions

ha0ye:
The definitions in the community blog feel more closely tied to the educational mission of The Carpentries, so I do like that aspect of those definitions.

At the same time, some details that I like about the existing definitions in the lesson, from the U. Pitt DEI glossary:
* the definition for `equity` specifically mentions `desirable outcomes` and not just participation and opportunities
* the definition for `inclusion` specifically mentions sharing power, which gives a little bit more insight into what is meant by "engagement"

kariljordan:
I really appreciate this issue and @ha0ye's feedback. I wonder if we can refine the glossary definitions to incorporate the sentiment of U. Pitt’s definitions while ensuring they clearly reflect what these terms mean within our community. I agree that reinforcing these definitions is important, as they shape how we make decisions, interact with each other, and uphold our shared values. I tried, but now these seem really long:)

- Equity: The creation of opportunities for access, participation, and the proportional distribution of desirable outcomes across groups. Equity focuses on addressing disparities and closing participation gaps in our community. While often confused with equality, which emphasizes equal treatment, equity is centered on achieving fair and just outcomes.

- Inclusion: The active, intentional, and ongoing engagement of traditionally excluded individuals and groups in processes, activities, and decisions in a way that shares power. Inclusion fosters broad participation, increases awareness and understanding, and cultivates a sense of belonging by creating safe, positive, and nurturing environments.

[...]

PR #1872, opened 2025-07-24
Repair source for DEI glossary terms

### LINKED ISSUE #1869: pitt glossary link is broken
### What is the problem?

broken link to source

### Location of problem (optional)

https://github.com/carpentries/instructor-training/blob/main/episodes/09-eia.md?plain=1#L328

brownsarahm:
We might want to use a different strategy for this though; the content was not moved so the "fix" is a webarchive link, but maybe we do not want to do that? 

closes #1869


kcranston:
I agree that a web archive link is not ideal, particularly in this context. Maybe wait to see the results of the discussion in #1797 before making a change to the reference link? If we do end up adding a reference or two here, having a few from outside the US might be wise.

PR #1875, opened 2025-07-28
warn that there are no reminders in signup exercise

brownsarahm:
responds to a trainee question

(if accepted this should also get updated on the [etherpad template](https://pad.carpentries.org/ttt-template))


kcranston (approved):

kcranston:
Who is responsible for maintaining the etherpad template? Is it us?

PR #1932, opened 2025-11-12
Updates to accessibility language

elletjies:
The term “accommodations” can have different translations across regions (i.e., accommodation = hotel room). To promote clarity and inclusivity, the Community Engagement Team decided to change all references from “accommodations” to “accessibility support” (adding that it is also referred to as accommodations)

Changes included: 

Accommodations → accessibility support
Accommodations request → accessibility support request
Accommodations request form → accessibility support request form.

### DIFF
--- a/episodes/09-eia.md
+++ b/episodes/09-eia.md
@@ -62,6 +62,14 @@ This exercise should take about 5 minutes.
 Barriers to accessibility encountered in a workshop are demotivating at best and can exclude a learner entirely.
 What barriers might be present in your workshop, and what can you do to remove them?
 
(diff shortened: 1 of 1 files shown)


ndporter:
I like the impetus here, but the language "accessibility support" also seems to kind of obscure the difference between accommodation (or other words that fit better in a non-US context) and universal design. That is, UD provides accessibility support, as do accommodations. But accommodations are specifically contrasted because they're reactive instead of proactive. I don't have a better choice to hand (though I'll try to think of one), but I think it does create some side effects in the section contrasting the 2 approaches.

PR #1955, opened 2025-12-15
Updated Welcome Session link

elletjies:
Moving forward all welcome session registration will take place in Pretix - https://pretix.carpentries.org/community-events/welcome-sessions/


kcranston:
Thanks for the updated link! The text before the link ("Visit the Welcome Session Etherpad") should also change, since it is no longer an etherpad.

PR #2024, opened 2026-07-17
Fix grammar in licensing statement for Carpentries lessons

somet-code:
Add missing works to the licensing statement to correct grammar


brownsarahm:
this change makes the sentence incorrect.

